PR #22, opened 2017-01-25
Updating 0007-Editors-API-Specification.md

singhsarab:
Adding the documentation for
- Cancel and Abort for Test Run
- Test Session messages
- Test Session termination


codito:
@singhsarab do we plan to add Debug protocol in a separate PR?

PR #31, opened 2017-03-01
Add TestCaseFilter documentation

smadala:
Related Issue: https://github.com/Microsoft/vstest/issues/334


AbhitejJohn:
Looks great. One point we might want to put in some place is that, we currently do not support a != on Categories AFAIR. I'm guessing we will touch on filtering in VS IDE in a separate doc file. (It isn't true filtering though)

PR #33, opened 2017-03-02
Initialize Extensions Docs

singhsarab:
Adding documentation for initializing extensions and setting the target framework

### DIFF
--- a/RFCs/0007-Editors-API-Specification.md
+++ b/RFCs/0007-Editors-API-Specification.md
@@ -111,6 +111,26 @@ At this point the initial handshake between Editor and Test Platform is
 complete. Above operation is required only once per launch of test platform.
 Subsequent discovery, execution operation don't require an initialization.
 
+### Initialize Extensions (Request)
+This request ensures that all the extensions that are required for discovery or execution
+get loaded before the actual operation request. This includes path to the test adapters.
+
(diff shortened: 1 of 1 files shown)


RFCs/0007-Editors-API-Specification.md
```diff
@@ -111,6 +111,26 @@ At this point the initial handshake between Editor and Test Platform is
 complete. Above operation is required only once per launch of test platform.
 Subsequent discovery, execution operation don't require an initialization.
 
+### Initialize Extensions (Request)
```
codito: Please add a note that this is required for `net4x` targets only.

Should we suggest the caller to pass output path of test project here? For most of the frameworks, test adapter is copied to output directory.

RFCs/0007-Editors-API-Specification.md
```diff
@@ -1264,6 +1284,39 @@ This request is used to the end the current test session.
 ```
 
 ## Appendix
+### Setting Target Framework
```
codito: Move this below **Key data structures**. Add a shortcut link `<a name="RunSettings.TargetFramework"></a>`. Link it from the test execution request section above.

For an example of linking see how `Message` (in Key Data Structures) is linked.

PR #59, opened 2017-06-30
Created RFC for CollectSourceInformation flag

Shyam-Gupta:


pvlakshm:
Are the following assumptions correct?
1. If there is no explicit .runsettings file associated, then it will be synthesized with the property ```CollectSourceInformation``` set appropriately.
2. If there was an explicit .runsettings file associated, and it did not contain this property, then the property would be added dynamically and set to the appropriate value.

What happens in the following cases?
1. What if there was an explicit .runsettings file associated which already had this property set to ```false```? Does that get overridden in some cases?
2. What happens if adapters ignore this property? There will be double-discovery of source information, yes? Will that result in any subsequent consistency issues?

Shyam-Gupta:
@pvlakshm:

Both the assumptions are correct.

Behavior in the two cases will be:
1. If an explicit .runsettings file has this property declared then it will not get overridden and will be passed as is to the adapter
2. In case adapter ignores this property and collects source information for the test methods then test explorer will not collect source information again and will honor the information provided by the adapter. It will not impact any subsequent scenario.

RFCs/0010-Source-Information-For-Discovered-Tests.md
```diff
@@ -0,0 +1,71 @@
...
+[$RootKey$\FeatureFlags\TestingTools\UnitTesting\UseSipToFetchSourceInformation]
+5. Update its dword value from 00000000 to 00000001
+6. Save the file and update it in the Visual Studio location mentioned in (2)
+7. Open Developer Command prompt in administrator mode and run following command:
+   devenv /UpdateConfiguration
+8. Launch Visual Studio and do test discovery for a managed (C# or VB) test project
+
+To disable the feature, restore backup file in the Visual Studio location mentioned in (2) and run step (7)
+
+Note that for perf improvement to show up, the test adapter needs to be updated to honor 'CollectSourceInformation' flag.
```
jayaranigarg: It will be good to mention the adapter version here which supports this flag i.e. 1.2.0-beta onwards.

Shyam-Gupta: Thanks for the suggestion. Doc updated with this information.

RFCs/0010-Source-Information-For-Discovered-Tests.md
```diff
@@ -0,0 +1,69 @@
...
+<RunSettings>
+    <RunConfiguration>
+      <CollectSourceInformation>False</CollectSourceInformation>
+    </RunConfiguration>
+</RunSettings>
+```
+
+Following are the perf numbers obtained for this flag in IDE scenario for MSTest and XUnit test adapters on TPv2 for a test project having 25K tests:
+
+| Adapter         | Without Changes = A (Sec) | With Changes = B (Sec) | Improvement, C = A - B (Sec) | Improvement (%) = C * 100 / A |
```
codito: Can we validate the perf benefits in nunit too? /cc: @jnm2 @rprouse @OsirisTerje

RFCs/0010-Source-Information-For-Discovered-Tests.md
```diff
@@ -0,0 +1,69 @@
...
+b. Performance improvement in CLI scenario:
+* Source information is not required in CLI scenario and can be skipped entirely
+
+## Overview of changes
+Visual Studio Test Explorer needs source information for discovered tests to power following 3 scenarios:
+1. Navigation to test method
+2. Test method execution from code editor context menu
+3. Code lens markers
+
+At present, for managed (C# and VB) test projects, this information is obtained from PDBs using reflection based APIs. While this approach works, it has perf impact.
```
AbhitejJohn: you might want to put this out separately so its more explicit.

PR #60, opened 2017-07-04
RFCs related to discovery via Test Explorer

singhsarab:


RFCs/0011-Discover-Only-TestProjects.md
```diff
@@ -0,0 +1,76 @@
+# 0011 Discover for Only Test Projects
```
codito: Test Projects Discovery in Test Explorer

RFCs/0011-Discover-Only-TestProjects.md
```diff
@@ -0,0 +1,76 @@
...
+for making improvements in the test discovery.
+
+## Motivation
+* Scope of improvement in discovery through TestExplorer. See [here](#Discovery.Proof)
+* Need to standardize the rules for identifying Test projects.
+
+## Problem
+TestWindow component inside VS IDE currently sends all the projects present in the solution for discovery.
+This list of projects includes projects that do not contain any tests.
+Sending redundant projects to the runner creates an overhead. Each adapter will try and fail to discover any tests inside such projects.
```
codito: Adapters have a reference check today to fast return if the project is not a test project. Still it has reflection cost.

RFCs/0011-Discover-Only-TestProjects.md
```diff
@@ -0,0 +1,76 @@
...
+
+### Scenarios
+There are two ways user can author C# based tests
+1. Create a test project using Unit Test template
+2. Add the appropriate adapter or framework references to a class library project. 
+
+### Current State
+There are existing ways to identify if a project is a test project. Brief explanation is as follows.
```
codito: nit: There are three mechanisms Test Explorer uses to identify test project. Number them below.

codito: What does Test Explorer do with the identified test projects?

RFCs/0011-Discover-Only-TestProjects.md
```diff
@@ -0,0 +1,76 @@
...
+### Discover All projects.
+All projects are sent for discovery operation and service guid will be added to projects if required.
+In short, experience is same as the existing one.
+This option would be selected by default.
+
+### Discover Test projects only.
+This would ensures only test projects are selected for discovery.
+* Note: Since we would not be sending other projects, the tests authored in projects that do not fulfil
+any of the rules will not be discovered. Here are a few options recommendations for users:
+1. Add *TestContainer* as a project capability.
```
codito: Shouldn't the adapters define this if they are participating in the build process?
Or, will Microsoft.NET.Test.Sdk define this?

codito: Provide an example how can an user manually define it.

RFCs/0011-Discover-Only-TestProjects.md
```diff
@@ -0,0 +1,76 @@
...
+2. Discover using All projects option, this would make sure service guid gets added to projects which contain
```
codito: In subsequent discovery, will test explorer send only the test projects?

RFCs/0011-Discover-Only-TestProjects.md
```diff
@@ -0,0 +1,76 @@
...
+tests but do not have TestContainer capabilities or Test ProjectTypeGuid.
+
+## Proof of concept <a name="Discovery.Proof"></a>
```
codito: Is there a timeline when this experience will show up for desktop projects?

RFCs/0011-Discover-Only-TestProjects.md
```diff
@@ -0,0 +1,76 @@
...
+<ItemGroup>
+    <ProjectCapability Include="TestContainer" />
+</ItemGroup>
+```
+
+## Proposed Changes
+Add an ability to identify the test projects using all the three rules mentioned above.
+Also, provide user an option to run discovery operation for test projects only.
+
+User will be provide with two options,
```
codito: Assume this will be an option in Tools | Options | Test option page.

RFCs/0011-Discover-Only-TestProjects.md
```diff
@@ -0,0 +1,76 @@
...
+1. Discover All projects.
+2. Discover Test projects
```
codito: How will discover test projects work for `C++`, `Python`, `NodeJS`, `Javascript` tests? Will there be any impact on `any code`?

RFCs/0012-Initialization-of-Adapters-And-Extensions.md
```diff
@@ -0,0 +1,36 @@
...
+User can upgrade the adapter nuget packages in the same VS session.
+In this case, older version of extension had already been initialized and then new extension is sent for
+initialization.
+
+### Multiple versions in the same solution
+User can have projects targeting different version of the test adapter.
+In case of net46 framework where host process is shared, the right adapter does not get initialized.
+
+## Proposed Changes
+Here are a few proposed changes in this regard:
```
codito: What are the `cons` of this approach? Any alternatives we have considered?

RFCs/0012-Initialization-of-Adapters-And-Extensions.md
```diff
@@ -0,0 +1,36 @@
...
+
+1. All the dependencies including test adapters are copied to the bin directory of the project.
+This is similar to setting /testAdapterPath to bin directory.
+2. Each test container will be hosted in a separate host. An option can be given to the user to choose
```
codito: Not sure if `shared` and `unshared` should be exposed to the user. Can you elaborate more on why is it an user concern?

RFCs/0012-Initialization-of-Adapters-And-Extensions.md
```diff
@@ -0,0 +1,36 @@
...
+between shared and unshared hosts.
+
+Note:
+* For C# based test, explicit initialization of adapters can be avoided, and hosts will pick the appropriate
+adapters from the bin directory.
+* In case of cli, user will still be able to use /testAdapterPath to initialize the extensions.
+* Having Unshared hosts will impact performance. <TODO: Add numbers>
+* Extensions shipping as VSIX will continue to be initialized.
```
codito: They will continue to be passed onto all the test hosts? Isn't that a performance concern?

Same is true for inbox adapters. Will they also continue to be provided to the test hosts?

RFCs/0012-Initialization-of-Adapters-And-Extensions.md
```diff
@@ -0,0 +1,36 @@
...
+
+1. All the dependencies including test adapters are copied to the bin directory of the project.
+This is similar to setting /testAdapterPath to bin directory.
+2. Each test container will be hosted in a separate host. An option can be given to the user to choose
+between shared and unshared hosts.
+
+Note:
+* For C# based test, explicit initialization of adapters can be avoided, and hosts will pick the appropriate
+adapters from the bin directory.
+* In case of cli, user will still be able to use /testAdapterPath to initialize the extensions.
```
codito: Is it an interesting scenario?

pvlakshm:
Just for my understanding ...
1. For projects containing tests that were not created using the test project template, the project file will be updated by inserting the service GUID post-discovery? Am I correct?
2. In the case where the project is under source control, there is an implicit expectation then that the user checks-in the updated project file? For the user working on just the tests, could it be that (s)he might not expect the project file to have got updated, and might exclude it from commits?
3. So, projects created using the test project template will have **(a)** the test project GUID **(b)** a service GUID **(c)** ```TestContainer``` as a capability, and **(d)** references to the relevant test framework?
3. How will this work for cases where there is no notion of a test project _per se_? Like JavaScript, etc.?
4. The experience change will need to be reconciled against other experience changes as appropriate.

RFCs/0011-Discover-Only-TestProjects.md
```diff
@@ -0,0 +1,76 @@
...
+* Note: Since we would not be sending other projects, the tests authored in projects that do not fulfil
+any of the rules will not be discovered. Here are a few options recommendations for users:
+1. Add *TestContainer* as a project capability.
+2. Discover using All projects option, this would make sure service guid gets added to projects which contain
+tests but do not have TestContainer capabilities or Test ProjectTypeGuid.
+
+## Proof of concept <a name="Discovery.Proof"></a>
+In Visual Studio 2017 15.3, all .net core tests projects are filtered based on TestContainer capability.
+Discovery operation with only filtered test projects is performant by 10-15% in large solutions.
+
```
navin22: @ : Should we explicitly tell, this is only for the csproj or dll based containers.

RFCs/0011-Discover-Only-TestProjects.md
ManishJayaswal: Should we add a build warning if project contains tests but does not have this capability in it's project file? This will be a breaking change though so we should do A/B testing and also provide very specific guidance on how to add the capability. We should also provide users with an ability to suppress this warning.
I think we should definitely do it for .NET Core projects and solutions which have mix of .NET Core and desktop projects - which should address majority of cases and those users are much more likely to make the required changes as their solutions would be new or under active development. We should collect more feedback before doing it for desktop only solutions.

ManishJayaswal: @srivatsn @davkean - do you guys have thoughts here?

davkean: Can we pull data on how many projects will be in this situation? Can we make changes to existing targets, etc to avoid it for most projects?

davkean: Adding a "TestContainer" capability should never be a recommendation for users - capabilities are not user settings. They should told to upgrade packages to ones that ship with the TestContainer capability, which I'm assuming we'll update xUnit, NUnit, etc?

RFCs/0012-Initialization-of-Adapters-And-Extensions.md
ManishJayaswal: There have been several requests to not have reference to adapters in sources as they are VS specific and makes no sense in open source development if the developer is not using VS. Has there been any thoughts on decoupling this from projects had have every solution pick from the adapters that come with VS or are on the machine? I understand that this is a completely different model but I was wondering if this will simplify things for future ( both for users and us) - i.e. you are ties to adapters available with VS on your box ( and not ones in your projects - ideally there should not be any in the projects)

ManishJayaswal: @tmat as he had at one point expressed opinion that we should not have adapters in sources and should use the ones that VS provides.

codito: (adding more history for context)

Today VS supports finding adapters from a vsix extension. So even if user doesn't have a adapter nuget package, the adapter from vsix could be used to run it. E.g. [here's](https://marketplace.visualstudio.com/items?itemName=NUnitDevelopers.NUnit3TestAdapter) the one for NUnit framework.

The challenge with this was adapters have a coupling with the test framework version. E.g. mstest v2 will not be able to run mstestv1 references. It was leading to hard to trace errors in such incompatibilities. Secondly, in CI machines, it required VS to be available with the exact same version of adapter installed as vsix. I think few test frameworks have now deprecated vsix adapter extensions.

Nuget packages were easier to maintain (same process as other dev dependencies), provided a way to use the same adapter both in CI and local dev box. But it led to scenarios with multiple adapters being sent for same project.

RFCs/0011-Discover-Only-TestProjects.md
singhsarab: Yes.

RFCs/0011-Discover-Only-TestProjects.md
singhsarab: Scenarios where TestExplorer is not pinned and no test projects get identified, Test related commands would not work from context.
This is because Test package gets loaded only if there are test projects in the solution.

RFCs/0011-Discover-Only-TestProjects.md
bradwilson: For xUnit, this is a file system check for both v1.x and v2.x artifacts.

RFCs/0011-Discover-Only-TestProjects.md
```diff
@@ -0,0 +1,76 @@
...
+* Scope of improvement in discovery through TestExplorer. See [here](#Discovery.Proof)
+* Need to standardize the rules for identifying Test projects.
+
+## Problem
+TestWindow component inside VS IDE currently sends all the projects present in the solution for discovery.
+This list of projects includes projects that do not contain any tests.
+Sending redundant projects to the runner creates an overhead. Each adapter will try and fail to discover any tests inside such projects.
+
+### Scenarios
+There are two ways user can author C# based tests
```
bradwilson: Just want to make sure you don't miss the fact that unit tests can be written in C#, VB.NET, and F# (with xUnit, and presumably NUnit as well).

RFCs/0011-Discover-Only-TestProjects.md
```diff
@@ -0,0 +1,76 @@
...
+1. Create a test project using Unit Test template
+2. Add the appropriate adapter or framework references to a class library project. 
+
+### Current State
+There are existing ways to identify if a project is a test project. Brief explanation is as follows.
+
+#### Test project Guid
+All C# based classic Unit tests which are created using the default templates have a special ProjectType guid to be identified as Test project.
```
bradwilson: I believe this only applies to MSTest.

RFCs/0011-Discover-Only-TestProjects.md
bradwilson: It's not clear to me that this is ever going to be a good idea, because it (presumably silently) blocks you from manually adding unit test support to a new or existing library; it will be ignored unless it contains one of the magic indicators. I personally can't ever imagine setting (or recommending to anybody) "Discover Test Projects". This feels like you're trying to solve a specific problem without naming it, and doing so in such a way that adding new test projects becomes potentially very user-unfriendly.

RFCs/0012-Initialization-of-Adapters-And-Extensions.md
bradwilson: VSIX-based test adapters were and are a terrible idea.

Using NuGet to express your dependencies explicitly on a per-project basis makes much more sense. You should never send an assembly to a test adapter which doesn't reference it. You're just wasting time.

The smart move would be to deprecate VSIX-based adapters entirely when moving to this system. That would make things much easier to reason about and much less bug prone.

RFCs/0011-Discover-Only-TestProjects.md
singhsarab: Yes. That's correct.

RFCs/0012-Initialization-of-Adapters-And-Extensions.md
singhsarab: Yes, for managed projects, what you are suggesting makes sense.

But there are scenarios, like in case of cpp and javascript tests, the project does not have a reference to test adapter, and has to be picked from the installed vsix extension path.
We are open to suggestions in this regard.

PR #77, opened 2017-09-16
Documented "prefix" option

harshjain2:
Added documentation for "prefix" option.

### DIFF
--- a/docs/report.md
+++ b/docs/report.md
@@ -95,9 +95,10 @@ Console logger is the default logger and it is used to output the test results i
 #### Syntax
 
 ```
-/logger:console[;verbosity=<Defaults to "minimal">]
+/logger:console[;verbosity=<Defaults to "minimal">;prefix=<Defaults to false>]
  
-Argument "verbosity" define the verbosity level of console logger. Allowed values for verbosity are "quiet", "minimal" and "normal".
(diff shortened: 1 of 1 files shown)


docs/report.md
```diff
@@ -95,14 +95,16 @@ Console logger is the default logger and it is used to output the test results i
 #### Syntax
 
 ```
-/logger:console[;verbosity=<Defaults to "minimal">]
+/logger:console[;verbosity=<Defaults to "minimal">;prefix=<Defaults to false>]
  
-Argument "verbosity" define the verbosity level of console logger. Allowed values for verbosity are "quiet", "minimal" and "normal".
+Argument "verbosity" defines the verbosity level of console logger. Allowed values for verbosity are "quiet", "minimal" and "normal".
+Argument "prefix" decides whether Verbose level should be added as prefix or not in log messages. Allowed values for prefix are "true" and "false".
```
Faizan2304: Should we explain to user what how prefix=true will change the output? I think it will make more clear to understand the use of prefix.

docs/report.md
```diff
@@ -95,14 +95,16 @@ Console logger is the default logger and it is used to output the test results i
...
 ```
 
 #### Example
 ```
 vstest.console.exe Tests.dll /logger:"console;verbosity=normal"
+vstest.console.exe Tests.dll /logger:"console;verbosity=normal;perfix=true"
```
jp2masa: Typo: `perfix` -> `prefix`.

PR #90, opened 2017-10-16
Update Code Coverage docs to point to released 15.3

twsouthwick:
The code coverage documentation was referencing a prerelease of 15.3, both for VS and the test SDK. This change updates the docs to point to the release versions.


codito:
Thanks @twsouthwick

PR #95, opened 2017-10-23
Guidelines for a TestCase's FullyQualifiedName

peterwald:


RFCs/0014-TestCase-FullyQualifiedName.md
```diff
@@ -0,0 +1,48 @@
...
+    Namespace.Namespace.Class.Method()
+    Namespace.Namespace.Class<Int32>.Method()
+    Namespace.Namespace.Class<Int32>.Method<String>()
+    Namespace.Namespace.Class+InnerClass.Method
+    Namespace.Namespace.Class.Method(x: 1, msg: "string")
+    Namespace.Namespace.Class.Method(1, "string")
+    Namespace.Namespace.Class(init: "Value").Method(param: 23)
+    Namespace.Namespace.Class<String>("Value").Method(23)
+
+## Normalized Form
```
AbhitejJohn: I'm guessing this would be set by the platform. Would /testcasefilter be using the normalized form for filtering? Or would users be required to pass in the FullyQualifiedName format above for filtering?

RFCs/0014-TestCase-FullyQualifiedName.md
```diff
@@ -0,0 +1,48 @@
...
+When referencing test cases by FullyQualifiedName, the test platform recognizes a normalized form of the FullyQualifiedName. This normalized form is the FullyQualifiedName with segments in parentheses and segments in angle brackets removed. The normalized form will not be unique among TestCases.
```
AbhitejJohn: I'm guessing we would be adding an extra property of NormalizedFQN on TestCase object. This would result in a json protocol change.

RFCs/0014-TestCase-FullyQualifiedName.md
```diff
@@ -0,0 +1,48 @@
...
+1. **FullyQualifiedName should be deterministic** FullyQualifiedName should have the same value for the same test across different discovery sessions. This means that including random data, time-dependent data or environment specific data in the FullyQualifiedName should be avoided.
+1. **FullyQualifiedName is expected to be a reference to the method that implements the test** The Test Platform is expecting the FullyQualifiedName to refer to a particular test case. If there are multiple tests that are instantiated from a test method (data-driven tests for instance), then the FullyQualifiedName should distinguish these using parentheses to represent data parameters and angle brackets to represent type parameters. See below.
+
+## Recommended Form
+FullyQualifiedName is expected to be a reference to the TestCase method that implements the test (or tests). The FullyQualifiedName should be unique to a particular TestCase instance, if possible.
```
AbhitejJohn: I'm wondering when FullyQualifiedName would be used and when would its normalized form be used. Is the below the right usage?
FQN:
1. Test Explorer Group By providers(Class/Namespace).
2. Default `TestCase.ID` generation.
3. Existing Filtering.

NFQN:
1. Fast Filtering.

navin22:
Tagging @codito

RFCs/0014-TestCase-FullyQualifiedName.md
navin22: NFQN will be set by the platform ? or Adapters needs to set?

peterwald: Yes, that's basically right. The main reason for creating the normalization rule is so that the TestWindow can filter tests without reproducing the FQN generation logic of each test framework. The normalization is currently implemented using the Test Window's replacement regex.

RFCs/0014-TestCase-FullyQualifiedName.md
peterwald: Currently this is handled by the TestWindow using the regex replacement logic. We did this so that there wouldn't be any requirement to change the Test Platform. There would be some benefits to doing this in the TestPlatform, but it's not an immediate requirement.

The main reason for this document is not to specify changes to the TestPlatform, but to document the normalization process for the benefit of test adapter developers.

RFCs/0014-TestCase-FullyQualifiedName.md
peterwald: I don't think we need to add a property yet, but it may be helpful in the future.

RFCs/0014-TestCase-FullyQualifiedName.md
peterwald: @Navin22 my previous comment was a response to @AbhitejJohn 

NFQN does not need to be set by the platform or the adapter. I am looking to document the process we are following in the VS TestWindow.

RFCs/0014-TestCase-FullyQualifiedName.md
```diff
@@ -0,0 +1,48 @@
...
+
+The recommended forms of FullyQualifiedName are described by the following examples:
```
codito: I think this form only applies to languages with NS.Class.Method concepts. Normalization will be skipped for languages that don't have these (e.g. Python or JS)?

RFCs/0014-TestCase-FullyQualifiedName.md
AbhitejJohn: Sounds good Peter. One consequence of this however could be that all adapters start putting in test data in their FQN from the guidelines above and that could break filtering using `/testcasefilter:FullyQualifiedName="N.C.M(x: 24)"` when the method is adorned by a Random attribute where the data to the method is indeterminate by design.
To avoid such scenarios we might just want to recommend stripping off the data in FQN putting it in `DisplayName` instead. Thoughs?

RFCs/0014-TestCase-FullyQualifiedName.md
peterwald: That's a great point. Normalization is only used for source-based test discovery. I'll add a comment clarifying that.

RFCs/0014-TestCase-FullyQualifiedName.md
peterwald: I think we will want to support test data in the FQN in the future, so that it will be possible to reference each individual test instance. In the case where non-deterministic data is included, I agree that it would not be possible to select that test by filter. I think I'd prefer to add NormalizedFQN as a filter property in that case to allow both scenarios.

RFCs/0014-TestCase-FullyQualifiedName.md
genlu: Not clear to me in what scenario having a NFQN property on TC would be beneficial. it seems the only reason we are interested in NFQN is so we can have a placeholder TC in test window, which can be used to run all related tests by filter.

Ultimately, we probably need to have another way to run tests by name, other than using TestCaseFilter (we use filter mostly because it does most things we wanted with only minimal change), say by specifying a list of `(string source, string[] fqns)` so we eliminate the ambiguity and have the ability to run just one input of a theory test such as `N.C.M(x: 24)`, plus a chance to further improve perf (for example, we can have a API that simply passing those source-fqns pairs to adapter and let it decide what to do, instead of the callback model currently utilized by filter)

AbhitejJohn: Sounds good.

OsirisTerje: Please see comments on NUnit PR https://github.com/nunit/nunit3-vs-adapter/pull/405#issuecomment-340079832 to implement a normalized FQN in the adapter.  This has unwanted sideeffects in the Test Explorer.  The display name must match the FQN.  
Further, the following warning is seen in Windows test output 
[10/27/2017 9:32:03 PM Warning] A test with the same name 'NUnit.VisualStudio.TestAdapter.Tests.TestDiscoveryTests(NUnit.VisualStudio.TestAdapter.Tests.Fakes.FakeDiscoveryContext).VerifyTestCaseIsFound("MockTest3","NUnit.Tests.Assemblies.MockTestFixture.MockTest3")' already exists. This test is not added to the test window.  I assume these are related to the same.

peterwald: @OsirisTerje The display name does not need to be the same as the FQN, but the FQN and the display name should have the same level of uniqueness (ideally completely unique for both). The issue in the referenced PR is that the ID being assigned to the TestCase is not unique (as a consequence of the FQN not being unique, and the Id not being otherwise set).

singhsarab:
@peterwald Let's bump up the number for this RFC and push this in, I am guessing we have already addressed the comments, right ?

peterwald:
@singhsarab I've bumped the number. You can go ahead and merge it.

PR #106, opened 2017-12-12
Update Code coverage requirements

kendrahavens:
Even on the command line, Code Coverage requires the machine to have VS Enterprise installed. We need these requirements to be more clear.


kendrahavens:
@pvlakshm Is this change okay to merge? Thank you.

PR #125, opened 2018-04-02
Adapter lookup based on user input

mayankbansal018:


RFCs/0020-User-Specified-TestAdapter-Lookup.md
```diff
@@ -0,0 +1,41 @@
+# 0020 User Specified Test Adapter Lookup
+
+## Summary
+This note outlines the proposed changes for lookup and initialization of the test adapters, based on User input.
+
+## Motivation
+
```
singhsarab: This is not the motivation.
Let's put this under another heading, "Current behavior".

RFCs/0020-User-Specified-TestAdapter-Lookup.md
```diff
@@ -0,0 +1,41 @@
...
+## Principles
+1. The adapter referenced in the project should be used for test execution
+2. Performance should not degrade.
+3. Adapter lookup logic is consistent across runs.
+4. Adapter lookup logic should be clear and consistent for both IDE/Editor and CLI runs.
+
+## Proposed Changes
+Here are the proposed changes based on the sources available for lookup of the adapters.
+User can restrict adapter lookup to only adapters close to source location, & those present in TestAdapterPath under RunSettings, via setting **UseSpecifedAdapterLocations** flag in RunSettings to **true**.
```
singhsarab: nit: UseSpecifiedAdapterLocations

I find above a little confusing, as it does not tell me that adapters in the test source directory will be used.
Should we call it UseLocalAdapters ?

How about we go for a SkipDefaultAdapters instead ?

RFCs/0020-User-Specified-TestAdapter-Lookup.md
```diff
@@ -0,0 +1,41 @@
...
+
+### Details for lookup
+
+* If the flag "UseSpecifedAdapterLocations" is set to true, then adapters only from source, & TestAdapterPaths would be picked.
+* If no adapters are found in above locations, then platform will use other adapters sent by IDE, which include adapters from nugets, & Vsix.
```
singhsarab: If we were to include the second point, we should say PreferSpecifiedAdapterLocations

RFCs/0020-User-Specified-TestAdapter-Lookup.md
```diff
@@ -0,0 +1,41 @@
+# 0020 User Specified Test Adapter Lookup
+
+## Summary
+This note outlines the proposed changes for lookup and initialization of the test adapters, based on User input.
+
+## Motivation
+
+As per the current adapter lookup model, when ever a Run/Discovery call is made from IDE to TestPlatform, the test adapter initialization happens in multiple phases listed below
+
+* InitializeExtensions(): In this call, the IDE passes list of all adapters it finds in current session, these include adapters referenced via nuget in entire solution, & adpaters acquired via Vsix(e.g. GoogleTestAdapter, BoostAdapters, etc..)
```
peterwald: nit: ~~adpaters~~ adapters

RFCs/0020-User-Specified-TestAdapter-Lookup.md
CharliePoole: It's a valid nit, IMO. The user has specified what adapters to use in TestAdapterPath and/or RunSettings. They will then wonder "Why do I have to say to use what is specified?" However "UseLocalAdapters is also confusing. Perhaps "UseOnlyLocalAdapters" or "ExcludeDefault(?)Adapters".

RFCs/0020-User-Specified-TestAdapter-Lookup.md
CharliePoole: IMO you should always prefer the specified adapters, without the need to say so. IMO again... the second point is unnecessary and confusing.

RFCs/0020-User-Specified-TestAdapter-Lookup.md
```diff
@@ -0,0 +1,41 @@
...
+* If no adapters are found in above locations, then platform will use other adapters sent by IDE, which include adapters from nugets, & Vsix.
+
+### Perf Improvements
+Restricting adapters via above logic also results in perf improvements, as it filters out unnecessary adapters which should be not used for a given test source, primarily for a managed test source, it filters our native adapters, as they are acquired only via Vsix, or from default **Extensions** folder.
+
+The improvement is however seen when a few tests are run, for e.g. when running two tests via RunSelected options from IDE, we see the time reduced from **0:00:01.4782809** to **0:00:01.1312048** in IDE Test Output pane 
+
+## Open Questions:
+* Should there be a UI option to set value UseSpecifedAdapterLocations, under Tool->Option->Test ?
+* Should we spawn a new testhost for each test source when above option is enabled?
```
CharliePoole: It would be nice for those of us who create adapters to have an understanding of the current rules for when a new testhost is spawned. 😸

RFCs/0020-User-Specified-TestAdapter-Lookup.md
ManishJayaswal: +1 to the nit. For flag name we should use "UseUserSpecifiedAdapters" and in the VS IDE UI we should use text explaining this option - "Use the test adapters directly referenced by test projects or ones specified in the RunSettings file"

•	Yes we should have an option ( a check box)  in the UI to enable this flag. This should be a per solution setting and only for C# and VB as other languages are shipping and plan to ship adapters as vsix in the box ( like C++ and TS). This box should be checked by default.
•	From and IDE perspective, we should not spawn a new host for each test assembly. 
•	If no adapters are found when the above option is selected then we should probably not fall back. Instead we should educate users to uncheck this box. This box should be checked by default though.

RFCs/0020-User-Specified-TestAdapter-Lookup.md
mayankbansal018: In the long run, from platform's perspective we want to be in a position, where adapters are *only* picked from where the test source. We already have achieved this for net core, & now we want to do the same for full clr, & native projects as well.

Without breaking down existing users, this is our first step to shed the legacy that platform has been carrying.

mayankbansal018:
@onovotny I suppose this approach at least solves the issues for "Run All" where additional tests were discovered, as IDE would send us adapters which could discover the tests. Since we are going to ignore adapter inputs from IDE, if this option is enabled, & as long as the adapters are not dropped close to source, the platform would not discover tests for such projects.

For discovery from source, this really doesn't help at all, as all the has been done in IDE, but we can certainly revisit that problem by maybe identifying such projects, & not running SBD on them.

RFCs/0020-User-Specified-TestAdapter-Lookup.md
mayankbansal018: We discussed this further, & everyone agrees to not have fallback. Will update the RFC.

onovotny:
To me, if this doesn't address SBD, it's incomplete.

peterwald:
@onovotny If we are able to move SBD into the adapters, then it should inherit the same behavior.

ManishJayaswal:
Today SBD looks for unit test references to determine if it should discover tests from a given project or not. We can easily add a check for the presence of adapter. However the thinking for not doing that was that there are a lot of projects out there which were not developed in VS but they still have tests. We would like those folks to use VS as their development tool so if they open those projects then we would show them all the tests in TE window even if we know we can't run them. The plan was to either give a nice message about adding the adapter when the user tries to run them or put some marker in the UI to indicate that the tests can't be run because adapter is not referenced. We just did not get around to doing that. Hopefully we can get that added in one of the future updates.

RFCs/0020-User-Specified-TestAdapter-Lookup.md
```diff
@@ -0,0 +1,53 @@
+# 0020 User Specified Test Adapter Lookup
+
+## Summary
+This note outlines the proposed changes for lookup and initialization of the test adapters, based on User input.
+
+### Current Adapter Lookup
+
+In the current model, when ever a Run/Discovery call is made from IDE to TestPlatform, the test adapter initialization happens in multiple phases listed below
+
+* InitializeExtensions(): In this call, the IDE passes list of all adapters it finds in current session, these include adapters referenced via nuget in entire solution, & adpaters acquired via Vsix(e.g. GoogleTestAdapter, BoostAdapters, etc..)
```
cltshivash: instead of IDE please use client (like Visual Studio IDE)

RFCs/0020-User-Specified-TestAdapter-Lookup.md
```diff
@@ -0,0 +1,53 @@
...
+3. Adapter lookup logic is consistent across runs.
+4. Adapter lookup logic should be clear and consistent for both IDE/Editor and CLI runs.
+
+## Proposed Changes
+User can restrict adapter lookup to only adapters close to source location, & those present in TestAdapterPath under RunSettings, via setting an option IDE to true.
+
+The option would be present at Tool->Option->Test.
+
+Once enabled IDE will modify the runsettings it passes to platform during discovery/execution, by adding **UseSpecifedAdapterLocations** flag in RunSettings, & setting it to **true**.
+
```
singhsarab: How does this setting affect the CLI flow ?

RFCs/0020-User-Specified-TestAdapter-Lookup.md
```diff
@@ -0,0 +1,53 @@
...
+Once adapters are gathered from all the possible locations, this list is passed to testhost process, which then loads all these adapters, & passes test source(s) to all the adapters.
+
+Following the above adapter look up logic, we pass down test sources to adapters which have no business running them, for e.g. passing a managed test source to a native adapter(GoogleTestAdapter, BoostAdapters, etc.), & vice-versa.
+
+## Motivation
+
+In .NetCore scenario, we only pick adapters from test Source location, this helps
+* Testplatform, as it knows where to look for adapters
+* Test Framework/Adapter writers, they know where to drop adapters so they are consumed by platform,& finally
```
singhsarab: Motivation : Select the correct adapter for the run. Most managed adapters are available in the test source directory, hence can be picked up from later for the test run.

Proposed Changes : Introduce a setting to pick adapter from the source location.

These points should come out as pros, #1 is redundant.

estrizhok:
I have two points to get you attention to with respect of the proposed changes:
1. It does not help the case where user references more than one UT framework in their project and tries to run only tests for one of them.
2. I also do not like the fact it does not directly let the client (IDE) choose what adapters will be initialized, i.e. regardless of the value of `UseSpecifedAdapterLocations`, test platform will always initialize all adapters from test source directory.

mayankbansal018:
@estrizhok Regarding your questions above:
1.  If the user is using multiple UT frameworks for a single test assembly, & runs a single tests, this approach won't help in such cases.

2. As I mentioned above in Motivation, the adapter lookup for fullclr projects is not clean, we allow adapters to be picked from various locations(Vsix, TestAdapterPath, DefaultExtensions). The aim of this model is to take first step towards the approach we use for net-core, where adapters are picked *only* from where  test source is. Here we are eliminating adapters the come from Default Location, & Vsix. 
We essentially want adapter owners of each frameworks to drop them close to test source.

estrizhok:
@mayankbansal018 thank you for reply, but

1. I know it won't help. That's what my comment was about. Are you guys going to address this case?
2. Essentially, what I as a client want to able to do is to run tests using test adapters specified by me. Current API does not provide this option. And unconditional lookup for test adapters in source directory does not help. If we're adding `do not use default adapters` flag, why not add another flag `do not use any adapters but explicitly specified`?

P.S. Unless I totally misunderstood who the proposed `UseSpecifedAdapterLocations` setting is for. RFC suggests that the setting is modified by the user in runsettings config. But it is ambiguous who the user is: the actual user or an IDE acting on his behalf. If that's the latter, then there is another discrepancy: IDE *needs* (as opposed to can) to modify something that should have come from the user. In this light we should review other settings available in runsettings config, and we see that the IDE has two ways to specify test adapters. One through `InitializeExtensions()` API call and the other using `TestAdaptersPaths` property. Which one should we prefer? And why do we need both?

CharliePoole:
It has always seemed to me a weakness in the API that the `.runsettings` file is both edited by the user and also changed by the IDE. An overall review seems like a great idea.

mayankbansal018:
@estrizhok , Once a test is discovered, each test is marked with Executor URI of the test adapter which discovered it. In cases where a single project is referring to multiple UT frameworks, we will load both the adapters, however once it tries to run the test, it will only invoke the adapter with maching executor uri of the test case which was required to run.

Regarding if we want to use extensions only provided by an IDE, is something we haven't explored. Reason being we want to keep the experience of adapter lookup common in both IDE, & CLI, & to do so as platform we wish to define what that model would be. The easiest experience we believe for user would be that platform implicitly picks the adapters that are close to their source. Also as mentioned above, once as a platform we can do so, then it makes it clear for adapter owners as well on where they should drop the adapters.
We already have this model for dotnet-core projects, we are trying to bring this for fullclr projects as well.

Regarding *UseSpecifedAdapterLocations*: We are still having this discussion internally on whether we need to make it part of runsettings or not. Since if a UI option in present in IDE, then based on that any client(IDE) can simply choose not to invoke InitializeExtensions call, hence not provide the adapters.

mayankbansal018:
@CharliePoole, I'll answer this based on VS IDE perspective, which always honors the runsettings provided by the user, but based on certain user actions, for e.g. with AnalyzeCodeCoverage, also appends CodeCoverage datacollector section, etc.. but at the same time preserving any input given by user like TestSessionTimeOut etc.
However even if any IDE choose to modify those settings further, from platform perspective it treats it just as an input which user wanted to give, & not IDE acting on behalf of it.

I hope I answered your question, please let me know if I misunderstood it.

mayankbansal018:
@CharliePoole The runsettings passed to platform is amalgamation of user inputs, for e.g. even in CodeCoverage scenario, the user's might not have passed any runsettings at all, but based on their gesture(AnalyzeCodeCoverage), IDE needs to add something to runsettings, to make platform aware that you have to enable CodeCoverage.

I suppose the main question here is that, is there any additional information that is put in runsettings which differs from User's actions. This is something every client(IDE) needs to ensure that they don't add any additional input from their end.

Also even in this proposed scenario, the "UseSpecifiedAdaptersLocations" would only be part of runsettings if users enables it via UI option.

PR #133, opened 2018-05-09
Update standard spec for FullyQualifiedName

peterwald:
cc: @cltshivash @ManishJayaswal @pvlakshm


RFCs/0017-TestCase-FullyQualifiedName.md
```diff
@@ -1,49 +1,97 @@
...
+`clr` – The first segment ("host" segment) specifies how to interpret the rest of the URI. The only valid value at the time of this writing is "clr" to reflect support for tests written in CLR managed languages. This portion can be expanded in the future to add support for non-managed languages such as Python, Javascript, C++, etc…
+
+`m | o` – `'m'` indicates a method form FQN, `'o'` indicates a custom test case identifier that the test explorer will not attempt to interpret.
+
+`{type}` - The `type` segment includes the fully specified type name in metadata format:
+
+    NamespaceA.NamespaceB.ClassName`1+InnerClass`2
+
+* The type name must be fully qualified (in the CLR sense), including its namespace. Any generic classes must also include an arity value using backtick notation (`# where # is the number of type arguments that the class requires).
+* Nested classes are appended with a '+' and must also include an arity if generic.
```
Pilchie: You should be more specific about the arity of nested types.  I forget the details, but I believe that they don't behave the way one may expect in reflection.

RFCs/0017-TestCase-FullyQualifiedName.md
```diff
@@ -1,49 +1,97 @@
...
+* There should be no extra whitespace included in the segment
+
+`{method}` - The `method` segment includes the fully specified method including the method name and a list of its parameter types inside parentheses separated by commas.
+
+    MethodName`2(ParamTypeA,ParamTypeB,…)
+
+* If the method accepts no parameters, then the parentheses should be omitted.
+* If the method is generic, then an arity must be specified (in the same way as the type segment).
```
Pilchie: Note: refelction uses a double ` (like ``) to represent generic methods.

RFCs/0017-TestCase-FullyQualifiedName.md
```diff
@@ -1,49 +1,97 @@
...
+* Insignificant whitespace should not be included in the segment.
+* The list of parameter types must be encoded in the same way that a type in the `{type}` segment of the FQN is encoded.
```
Pilchie: How should instantiated generics be encoded?  For example, you could have two test cases, one with a parameter type of `List<int>` and another with `List<string>` and I don't see how you will distinguish them in this spec.

RFCs/0017-TestCase-FullyQualifiedName.md
```diff
@@ -1,49 +1,97 @@
...
+
```
Pilchie: Consider documenting how special names (overloaded operators, constructors, finalizers, etc should be named).

RFCs/0017-TestCase-FullyQualifiedName.md
```diff
@@ -1,49 +1,97 @@
...
+## Custom Test Naming
```
Pilchie: Consider documenting what to do with explicit interface implementation methods.

Pilchie: Note that these might contain instantiated generics as well (if I am implementing `List<int>`)

RFCs/0017-TestCase-FullyQualifiedName.md
```diff
@@ -1,49 +1,97 @@
...
+
+## Managed Tests that are Not Type/Method Based
+
+It is possible to create unit tests in managed code where tests are not necessarily based on types & methods. While this is somewhat uncommon, we need to be able to handle this situation gracefully. Some examples are gherkin-dotnet and custom discoverers in xUnit. In these examples, an externally parsed definition file may determine the tests. For these cases, we have defined the 'other' form of FQN. FQNs in this form will look similar to the following.
```
SabotageAndi: gherkin-dotnet is only the parser for Gherkin, which is used by SpecFlow. At SpecFlow we are generating coded tests in a code behind file (for xUnit, NUnit, MSTest or SpecFlow+Runner). The SpecFlow+Runner has a custom test discovery where not everything is only from the code, but also from configuration.

RFCs/0017-TestCase-FullyQualifiedName.md
peterwald: @Pilchie I haven't been able to find a scenario that doesn't work this way with regard to nested types. 

```csharp
public class A<T>
{
    // Named by reflection as A`1+B`1
    public class B<X> { }

    // Named by reflection as A`1+B`2
    public class B<Y, Z> { }
}
```

Is there another scenario we should be considering?

RFCs/0017-TestCase-FullyQualifiedName.md
peterwald: From what I can see in the reflection APIs, the generic type names are actually used in the name. I don't see double ` appearing anywhere.

```csharp
        public class B<X>
        {
            // Reflection shows as Void MethB[V](X, V)
            void MethB<V>(X x, V v) { }
            // Reflection shows as Void MethB[V,Z](X, V, Z)
            void MethB<V,Z>(X x, V v, Z z) { }
        }
```

RFCs/0017-TestCase-FullyQualifiedName.md
peterwald: This is a good point. The parameter types need to be the instantiated types, rather than the generic types. I'll clarify that. I expect that using angle brackets to denote this would probably be the most straightforward.

RFCs/0017-TestCase-FullyQualifiedName.md
peterwald: Yes, this is worth expanding. Although I know the latest version of xUnit prohibits using explicit interface implementation methods as tests, there may be another framework that does permit it.

RFCs/0017-TestCase-FullyQualifiedName.md
peterwald: I'm not sure how useful a test method could be if defined as an operator or finalizer, but constructors might have some potential. It makes sense to address it.

RFCs/0017-TestCase-FullyQualifiedName.md
Pilchie: I think the issue I had in mind is just that reflection considers "T" to be a type parameter of B in a lot of ways.  Probably not an issue for the name.

RFCs/0017-TestCase-FullyQualifiedName.md
```diff
@@ -1,49 +1,97 @@
...
-| `Namespace.Namespace.Class(init: "Value").Method(param: 23)` | `Namespace.Namespace.Class.Method` |
-| `Namespace.Namespace.Class<String>("Value").Method(23)` | `Namespace.Namespace.Class.Method` |
+* `FullyQualifiedName` should uniquely identify a method that represents test cases (with the optional application of a dataset). There may be duplicate FQNs within a project due to some methods being associated with multiple test cases. Note that it is possible for multiple assemblies, or multiple test executors to use the same FQNs. When comparing FQNs within a Visual Studio session, `Source`, `FQN`, and `ExecutorURI` must be combined to guarantee uniqueness.
+* Each TestCase has an `ID` field that takes the form of a GUID. In the past, this field has been optional. It will now become required. The required `ID` field must be unique for a given test assembly. This `ID` field must also be deterministic, in the sense that a test with the same class, method and method arguments should have the same id. We envision generating this `ID` by hashing the test's class, method, method signature and method arguments into the GUID.
```
shyamnamboodiripad: The spec should probably clarify whether ID is unique across a given project or whether it can be unique across the solution i.e. whether or not TestCase.Source (and  TestCase.ExecutorUri) are included in the ID generation.

Personally I think having unique IDs across the solution would be great because that means any code that needs to keep track of TestCases can just hold them in a dictionary with ID as key. I just fixed an issue in LUT where the code was incorrectly assuming that it is safe to do this and fails because currently this not safe. It would be great if all code consuming TestCases could just use the ID as key without having to create an intermediate identifier type (comprised of ID, Source and ExecutroUri) for this. This would also make all code that deals with large dictionary of tests (the test store in TW, similar stores in SBD, LUT etc.) more efficient because storage and retrieval of TestCases can be performed by comparing just a single GUID (instead of comparing a GUID and two other strings).

RFCs/0017-TestCase-FullyQualifiedName.md
```diff
@@ -1,49 +1,97 @@
...
+Also, when working with unmanaged test adapters, we will need to handle the case of not having a standardized FQN. One way to support non-conforming adapters is to define an FQN format specifically for them so that we can still have them "standardized" without interpreting the contents.
+
+**Example**
+
+    fqn://compat/Class::Method::Test()
+
+## TestCase FQN Property Implementation
+
+We intend for test adapter implementors to use this new FQN format for the `FullyQualifiedName` property that is supplied to the TestCase object. Even though it is not specified, there may be some compatibility concerns with existing tools that depend on the `FullyQualifiedName` being in the current format. For an intermediate period, while compatibility concerns are worked through, we would suggest providing the new FQN in an additional custom property on the TestCase object named "StdFullyQualifiedName". Adding this additional property will affect the performance and allocation profile of the testing infrastructure, and we recommend not using this technique unless it is necessary for compatibility.
+
```
shyamnamboodiripad: The spec should probably also clarify whether a new field is needed for the new ID. I am guessing that TestCase discoverers (SBD, adapters) can just change the existing ID field to adhere to this spec. But I'm not sure if there are any compat concerns around that...

RFCs/0017-TestCase-FullyQualifiedName.md
shyamnamboodiripad: I think Artur had done some investigation around operators etc. to see if we need to worry about them for source based discovery.Iirc at least some of the frameworks allow such cases with no errors.

At the very least we should be able to handle such tests without crashing (even if we can't fully adhere to the spec in such cases). It would be nice to add unit tests for this (I can look into this).

RFCs/0017-TestCase-FullyQualifiedName.md
shyamnamboodiripad: You had alluded to using !0 and !!0 to encode type and method level generic parameters that appear in method signatures. Is this something that needs to be mentioned in the spec?

RFCs/0017-TestCase-FullyQualifiedName.md
shyamnamboodiripad: Need to call out the following -
- pointer types get *
- dynamic is replaced with object (the reflection-based implementation would only see object but we need to call this out for the Roslyn-based implementation for SBD)
- add examples for !0 and !!0 (including cases where there are generic params on outer and inner classes in a nested scenario)

shyamnamboodiripad: Also need to call out how arrays are handled (array of array, multi-dim arrays) since methods could differ just by the dimensionality of the array...

RFCs/0017-TestCase-FullyQualifiedName.md
shyamnamboodiripad: Per recent discussion, we could probably make things work with a single field. If so, we should update this portion of the spec accordingly...

RFCs/0017-TestCase-FullyQualifiedName.md
peterwald: True, that the nested classes include the template parameters of the containing class. So, nested class B<X> functions as if it were declared B<T, X> with a silent first template argument. We are numbering the template arguments including those that are received from the containing class. This is consistent with C# and the CLR metadata spec. I think we are on the same page here.

RFCs/0017-TestCase-FullyQualifiedName.md
peterwald: In the CLR metadata spec, generic class parameters are declared with '!' and generic method parameters are declared with '!!'. This might be what you are thinking of. I'm not aware of any case where double backticks are used in the method name. I think our usage is consistent here with all of the specs we've seen.

RFCs/0017-TestCase-FullyQualifiedName.md
```diff
@@ -1,49 +1,132 @@
...
+* `FullyQualifiedName` should reference the semantic location of the method that defines the test. This means that the `FullyQualifiedName` references the method on the type where the test is discovered. In some situations, this could be different from the type where the method is defined. See later in this document for a more detailed discussion.
 
+## `FullyQualifiedName` specification
 
-The normalized form allows the test platform to reference groups of tests in a more general form without having to interpret the logic that the test adapter uses when constructing the fully qualified name. 
+We are standardizing the `FullyQualifiedName` property on a TestCase in the following way. The FQN will be parsed as a URI starting with the scheme value of `fqn`. 
+
+**Method Form**
+
+    fqn://clr/m/{type}/{method}
```
abhishkk: There is a Test Plan/Test Suite scenario in vsts which requires [association](https://docs.microsoft.com/en-us/vsts/test/associate-automated-test-with-test-case?view=vsts) of test method. Associations stores FQN.

1. Currently we support association for mstest v1 adapter only. Changing the FQN for mstest v1 adapter will break all these existing TestPlan/TestSuite scenarios.
2. In addition, there is a huge ask from community to enable association for all adapters (including mstest v2, XUnit, NUnit). Once association is enabled for mstest v2 adapter, all the existing TestPlan scenarios which points to mstestv1 will not be able to migrate to mstest v2 (if this FQN change happens in mstest v2).

RFCs/0017-TestCase-FullyQualifiedName.md
```diff
@@ -1,49 +1,132 @@
...
+This note standardizes the form of `FullyQualifiedName` that test adapters should use when setting the `FullyQualifiedName` on test cases before returning them from the discovery process.
 
 ## Overview
-There are a broad variety of choices that current Test Adapter implementers have made when setting the FullyQualifiedName on TestCase objects created during test discovery. This note lays out some basic guidelines to make the usage of FullyQualifiedName more useful across adapters. These guidelines are focused on languages that support namespaces, classes and methods for structuring test cases (e.g. C# & VB). Other languages may not fit these guidelines.
+There are a broad variety of choices that current Test Adapter implementers have made when setting the `FullyQualifiedName` on TestCase objects created during test discovery. By standardizing the format of `FullyQualifiedName`, components in the vstest infrastructure can be implemented more reliably. These guidelines are focused on unit tests written in managed languages (e.g. C# & VB). Other languages may derive less benefit from adopting these guidelines.
```
abhishkk: Existing RFC for FQN talks about the recommended FQN form and the corresponding normalized form. Since this PR also focuses on managed languages only, what are the additional problems this PR solves?

I can see few additional features which this PR will introduce like
1. Better extensibility for languages other than managed, 
2. Generalized FQN across adapters.

But apart from feature additions, what existing problems this PR solves which current FQN RFC is not able to? This proposed RFC causes lot of compat issues. Please check below comments for details.

RFCs/0017-TestCase-FullyQualifiedName.md
```diff
@@ -1,49 +1,132 @@
...
+
+## Legacy Compatibility
+
+We intend to implement a shared routine that can convert a legacy FQN into this standardized format. Throughout the TestWindow code, in all other locations, we would use the new standardized format. Any edge cases in interpretation can likely be addressed in the conversion routine. This does mean, however, that when handling test cases from adapters that do not support the format, we will need to allocate another property to save the original legacy FQN value. This is necessary so that when we call back into the adapter, we supply the original FQN rather than the new one.
```
abhishkk: What are the compat plans for older VS with new adapter versions (containing new FQN)

RFCs/0017-TestCase-FullyQualifiedName.md
```diff
@@ -1,49 +1,132 @@
...
+* `FullyQualifiedName` should be immutable in that the test store keeps the exact value that is provided by the test adapter and supplies it back to the test adapter for test execution. The test adapter should be able to rely on receiving the same FQN for execution that it supplied when performing discovery.
+* `FullyQualifiedName` should reference the semantic location of the method that defines the test. This means that the `FullyQualifiedName` references the method on the type where the test is discovered. In some situations, this could be different from the type where the method is defined. See later in this document for a more detailed discussion.
 
+## `FullyQualifiedName` specification
 
-The normalized form allows the test platform to reference groups of tests in a more general form without having to interpret the logic that the test adapter uses when constructing the fully qualified name. 
+We are standardizing the `FullyQualifiedName` property on a TestCase in the following way. The FQN will be parsed as a URI starting with the scheme value of `fqn`.
```
abhishkk: Is there a plan to change the ObjectModel schema since FQN is proposed to be an URI instead of string?

RFCs/0017-TestCase-FullyQualifiedName.md
```diff
@@ -1,49 +1,132 @@
...
+## Background
 
-## Recommended Form
-FullyQualifiedName is expected to be a reference to the TestCase method that implements the test (or tests). The FullyQualifiedName should be unique to a particular TestCase instance, if possible.
+`FullyQualifiedName` (also referenced as FQN) is an identifier supplied by each VS test adapter for each test during discovery. That same FQN value is supplied back to the adapter (along with other TestCase properties) when it's time to execute the tests. One exception to this case is the "Run All" operation which discovers all tests and executes them in one operation. There has historically been no standard for this value and each adapter has implemented its own strategy for creating FQNs.
 
-The recommended forms of FullyQualifiedName are described by the following examples:
+There is a need in VS for each test to have a unique and deterministic method of identifying tests irrespective of the underlying test framework. A standardized FQN opens the path for a full and complete implementation of features like source-based discovery, dynamically fetching source information and UI elements like the hierarchy view in the Test Explorer window.
```
abhishkk: For all above features, why can't we let adapters to provide these info? Similar to ITestContainerDiscoverer, we can have one more extensibility point for adapters which can be used by TestWindow to query these info.

RFCs/0017-TestCase-FullyQualifiedName.md
peterwald: There is nothing wrong with storing the standardized version of the FQN. It is just a string that identifies a test case.

An adapter can tell the difference between an existing FQN and the standardized version by looking to see if the string starts with `fqn://`. So, it would know whether the FQN is standardized or not and would be able to adjust as needed. We also have a standard for representing the older FQN as a standardized FQN by the 'other' form (i.e. `fqn://clr/o/{non-standard fqn}`). So, if necessary, the non-standard FQN could be up-converted to a standardized form without interpreting it.

RFCs/0017-TestCase-FullyQualifiedName.md
peterwald: 1. For managed languages, it allows us to precisely find a particular method associated with a test case. This provides a few performance benefits...
    1. We can relate tests to source code without needing any additional information outside of the FQN. This means we don't have to incur the cost of PDB processing, and it provides performance benefits in the IDE.
    1. For fast execution of a few tests (out of a large test set), we can use the FQN for execution-by-name. The standardized FQN has enough information to select the test without performing a full discovery. If we know that a method is a test method, then it even allows test execution without any discovery at all. This is helpful for Live Unit Testing as well as the IDE.

This spec would replace the need for normalization. Normalization is not a sufficient solution for the following reasons.
1. Because it is implemented using test filters, it must use RegExp processing. Regular expressions do not support recursion and so there are some common cases that cannot be properly handled, such as nested parenthetical expressions, or nested generics. In addition, string parameters cause issues. We have run into bugs in customer scenarios and we cannot fix them by only performing RegExp filtering.
2. There are a few cases, such as explicit interface implementation, where the format of the existing FQN (and normalized form as well) cannot distinguish between the type's namespace and the method's interface. This is a result of using periods to separate the segments.

RFCs/0017-TestCase-FullyQualifiedName.md
peterwald: We plan to have a run setting that the new version of VS adds. This will tell the adapter to use the standardized FQN implementation.

RFCs/0017-TestCase-FullyQualifiedName.md
peterwald: No change to the schema. FQN is in the form of a URI, but it is still a string.

RFCs/0017-TestCase-FullyQualifiedName.md
peterwald: I'm not sure what you have in mind regarding a new extensibility point. Can you clarify? We are trying to streamline the interface with the adapters by reducing the data and processing required.

RFCs/0017-TestCase-FullyQualifiedName.md
abhishkk: 'Legacy compatibility' section of the PR mentions that how TestWindow will handle the scenario when adapters does not specify the proposed standardized FQN. 
It doesn't give clarity that adapter also needs to convert old FQN to standardized format.

So apart from moving to the new standard FQN, is there an expectation from adapters to have a compat mechanism as well (in which adapters converts the old FQN to new FQN)? Can you please update the RFC if that's the case?

This compat conversion seems more like the job of TestWindow or TestPlatform. Adapter should not be doing such compat conversions.

RFCs/0017-TestCase-FullyQualifiedName.md
abhishkk: Can you please update the RFC regarding the same?

1. So while discovery, adapter will create new FQN when runsettings has some particular value but old FQN when that value is not present?
2. And while execution, adapter will check if the FQN is old or new one. If old, convert it to new standard FQN and then run the tests?
2. What happens with vstest task? Does task flow will always go with old FQN flow OR user needs to manually update value in runsettings?

RFCs/0017-TestCase-FullyQualifiedName.md
abhishkk: `NOTE: Our intent is that the test adapters take responsibility for source-based test discovery. Once this transition is complete, the need to keep the FQNs in sync between discovery methods will be reduced to only making sure the types and methods match up.`

Sorry for confusion. My question is related to this section. Why not do this work in first place instead of generalizing FQN?

Instead of generalizing all adapters to follow a unique standard of FQN, why not let adapters make aware of the issues/improvements which they need to do for fast/reliable discovery and execution?
Few of the issues/improvements which adapters needs to be take care of are:
1. Source based discovery
2. Eliminating the need of discovery (even source based discovery) while execution by setting method info while discovery itself (this info can be set in custom properties or in FQN like this PR)

RFCs/0017-TestCase-FullyQualifiedName.md
SabotageAndi: @peterwald sorry for the late respone. The notification went under.
I think I only wanted to point out, that gherkin-dotnet is not "correct" here, as it is only the parser.

About the method-based FQNs & SpecFlow+Runner: I worked some weeks ago on an implementation of it and it looked good that we also can provide them. And as said, we always can fallback to the "other" method.

SabotageAndi:
As this RFC is still in writing phase, I assume the new FQN support will not be ready for VS 15.8.
Is there already a plan in which future VS version it will be then supported, so I can plan when to add it to the SpecFlow+Runner?
Thanks!

pvlakshm:
This will not be ready for VS 15.8.
This is still under review as you see, and we just want to ensure that it does not break any existing scenarios.
At the time this is reviewed and merged, will call out the target release as well.

SabotageAndi:
@pvlakshm Cool! Thanks for the info!

RFCs/0017-TestCase-FullyQualifiedName.md
peterwald: I've added an example to demonstrate this.

RFCs/0017-TestCase-FullyQualifiedName.md
peterwald: I agree this could make the implementation easier in certain ways. I have some concerns about this though.

1. Making the ID globally unique would require that the adapter & framework know all of the data required to make that ID unique including the Source and the ExecutorUri. Given that discovery is abstracted and implemented by the test framework core, does the discovery engine always have access to the source (the assembly file name) and executor? I'm not sure if it would in every case. What about scenarios that are not based on assemblies?
1. In the future, if additional scenarios are added that would allow for a single test case to have multiple "instances" within a single source assembly, it might be very difficult to support if we also require the test frameworks to know the scope of uniqueness of each test case.

It seems more consistent to scope the uniqueness guarantee of test case ids to be within the discovery container. We can always add another deterministic hashing layer at the TestWindow level that further scopes the ids by Source/TFM, and Executor, etc...

RFCs/0017-TestCase-FullyQualifiedName.md
peterwald: I'll add the explanation regarding the runsetting.

RFCs/0017-TestCase-FullyQualifiedName.md
peterwald: I've updated the document to clarify the run setting.

> So while discovery, adapter will create new FQN when runsettings has some particular value but old FQN when that value is not present?

Yes, the adapter can change whether it uses the std FQN or not based on the run setting. Clarified in the doc.

> And while execution, adapter will check if the FQN is old or new one. If old, convert it to new standard FQN and then run the tests?

During execution, the adapter should only be given the FQN that it created when doing discovery. If discovery and execution are happening across different versions then some one-time conversion might be required.

> What happens with vstest task? Does task flow will always go with old FQN flow OR user needs to manually update value in runsettings?

The user will have the option to update the runsettings, but I would expect that we would update the vstest task to provide the runsetting as soon as we complete the support.

RFCs/0017-TestCase-FullyQualifiedName.md
peterwald: This standardization has benefits in making the communication protocol between VS and the adapters more explicit. This has benefits in source-based discovery and execution-by-name as mentioned earlier, but it also creates a foundation for functionality in the future.

This means that we can have a robust foundation for the future, rather than continuing to layer fixes on top of fixes.

RFCs/0017-TestCase-FullyQualifiedName.md
peterwald: No, we don't expect adapters to need to convert the old FQN to the new FQN. We are only asking that the adapter switch to the new format. The runsetting will be provided to allow for handling compat issues, where the adapter would like to be aware if Std FQN is supported.

RFCs/0017-TestCase-FullyQualifiedName.md
abhishkk: >This has benefits in source-based discovery and execution-by-name
Thats what my comment is about. In this PR, we are asking adapters to move to standardized FQN so that source based discovery and execution by name can happen properly. Instead we should ask adapters to solve the source based discovery and execution by name problem on their own (its upon adapters to solve this problem via FQN or via some custom property or however they want to).

RFCs/0017-TestCase-FullyQualifiedName.md
abhishkk: >The user will have the option to update the runsettings

This value is not effecting any user experience. So, I don't think user should be aware of runsettings value which effects FQN.

>but I would expect that we would update the vstest task to provide the runsetting as soon as we complete the support.

We have already shipped TFS in which this update in task is not possible. Back compat will not be possible there.

RFCs/0017-Managed-TestCase-Properties.md
```diff
@@ -0,0 +1,109 @@
...
+* If the method is generic, then an arity must be specified (in the same way as the type property).
+* There must be no whitespace included in the segment
+* The list of parameter types must be encoded using the type specification below.
+* Return types are not encoded.
+
+### Parameter Type Encoding
+
+Parameters are encoded as a comma-separated list of strings in parentheses at the end of the `ManagedMethod` property. Each parameter is encoded using the rules below.
+
+* Basic Types - Types should be written using their namespace and type name with no extra whitespace. For example (`NamespaceA.NamespaceB.Class`). Native types should be written using their CLR type names (`System.Int32`,`System.String`). Return types are not encoded.
```
shyamnamboodiripad: The Return types are not encoded part seems redundant (its already mentioned in a bullet point above).

RFCs/0017-Managed-TestCase-Properties.md
```diff
@@ -0,0 +1,109 @@
...
+* Array Types - Arrays should be encoded as the element type followed by square brackets (`[]`). Multidimensional arrays are encoded the same as single dimensional arrays.
```
shyamnamboodiripad: Shouldn't multidimensional arrays specify arity via `,` inside the `[]`?

peterwald: Good catch, I've updated the doc.

RFCs/0017-Managed-TestCase-Properties.md
peterwald: Removed.

peterwald:
@pvlakshm This PR has been implemented for tests discovered in assemblies (Reflection) and source (Roslyn) with no modifications expected. Can we get this merged? Thanks.

PR #138, opened 2018-05-31
RFC

kaadhina:


RFCs/0023-TestSettings-Deprecation.md
```diff
@@ -0,0 +1,31 @@
+# 0023 TestSettings Deprecation
+
+## Summary
+
+This note outlines the proposed changes to deprecate support for testsettings in test runs.
```
PBoraMSFT: Should be "to deprecate support for using .testsettings in automated unit and functional testing scenarios."
We should also add that .testsettings itself will continue to be supported for load test scenarios.

RFCs/0023-TestSettings-Deprecation.md
```diff
@@ -0,0 +1,31 @@
...
+
+## Motivation
+
+Today there are two types of files to configure tests: *.testsettings and *.rusnettings. To simplify the experience, we are planning to deprecate support for *.testsettings.
```
PBoraMSFT: Wherever you say deprecate support, please qualify with 'for automated unit and functional testing scenarios'.

RFCs/0023-TestSettings-Deprecation.md
```diff
@@ -0,0 +1,31 @@
...
+|/TestSettings/Execution/TestTypeSpecific/UnitTestRunConfig           |/RunSettings/LegacySettings/Execution/TestTypeSpecific/UnitTestRunConfig    |
+|/TestSettings/Execution/Timeouts: testTimeout                        |/RunSettings/LegacySettings/Execution/Timeouts: testTimeout                 |
+|/TestSettings/Execution/Timeouts: runTimeout                         |/RunSettings/RunConfiguration/TestSessionTimeout                            |
+|/TestSettings/Execution: parallelTestCount                           |/RunSettings/LegacySettings/Execution: parallelTestCount                    |
+|/TestSettings/Execution: hostProcessPlatform                         |/RunSettings/LegacySettings/Execution: hostProcessPlatform                  |
+|/TestSettings/Execution/Hosts                                        |/RunSettings/LegacySettings/Execution/Hosts                                 |
+|/TestSettings/Execution/TestTypeSpecific/WebTestRunConfiguration     |/RunSettings/WebTestRunConfiguration                                        |
+
+You can migrate your existing testsettings file to runsettings file using a migrator tool that will ship with test platform.
```
singhsarab: Put this under a separate heading.
We should probably add details of migrator and how to use it.

RFCs/0023-TestSettings-Deprecation.md
```diff
@@ -0,0 +1,31 @@
...
+## Summary
+
+This note outlines the proposed changes to deprecate support for testsettings in test runs.
+
+## Motivation
+
+Today there are two types of files to configure tests: *.testsettings and *.rusnettings. To simplify the experience, we are planning to deprecate support for *.testsettings.
+
+## RoadMap
+
```
singhsarab: This might increase the scope of this RFC.

LegacySettings is not something we want to encourage as such. Ideally, we want users to move away from testsettings and use first class supported runsettings node.
Legacy should be used only if they are not able to achieve that using the existing settings.

For example, for parallel, user should actually try MaxCPUCount or/and MSTest parallelize instead of simply using legacy.

RFCs/0023-TestSettings-Deprecation.md
singhsarab: Nit: runsettings*

PBoraMSFT:
Looks good. Approved.

PR #141, opened 2018-06-06
Add escape sequence to filter doc

genlu:
FYI @smadala

### DIFF
--- a/docs/filter.md
+++ b/docs/filter.md
@@ -26,6 +26,19 @@ Allowed **operators**:
 
 **Value** is a string. All the lookups are case insensitive.
 
+**Escape Sequences** must be used to represent characters in the value that have special meanings in filter, i.e. filter operators.
+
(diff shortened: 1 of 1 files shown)


docs/filter.md
```diff
@@ -26,6 +26,19 @@ Allowed **operators**:
...
+| -------------- | -------------------- |
+| \\\\            |  \\                   |
+| \\(             |  (                   |
+| \\)             |  )                   |
+| \\&             |  &                   |
+| \\\|             |  \|                   |
+| \\=             |  =                   |
+| \\!             |  !                   |
+| \\~             |  ~                   |
+
```
smadala: Should we document [utility method](https://github.com/Microsoft/vstest/blob/master/src/Microsoft.TestPlatform.ObjectModel/Utilities/FilterHelper.cs#L23) available for clients here?

PR #152, opened 2018-08-31
LegacySettings

singhsarab:
Update runsetting with LegacySettings.

### DIFF
--- a/docs/configure.md
+++ b/docs/configure.md
@@ -56,6 +56,7 @@ The `runsettings` file is a xml file with following sections:
 2. Data Collection
 3. Runtime Parameters
 4. Adapter Configuration
+5. Legacy Settings
 
 We will cover these sections in detail later in the document. Let's discuss few
 core principles for runsettings.
@@ -72,6 +73,8 @@ core principles for runsettings.
    test run via the test platform APIs. It may use it to read its own settings
    or take decisions (e.g. an adapter needs to disable appdomains if
    `DisableAppDomain` setting is provided via run settings).
+5. Users can now configure settings used for legacy scenarios like deployment,
+   scripts using run settings.
 
 ### Sample
 Given below is a complete `runsettings` file with all available options. Each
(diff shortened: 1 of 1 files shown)


docs/configure.md
```diff
@@ -152,6 +155,47 @@ document.
...
   </InProcDataCollectionRunSettings>
 
+  <!-- Adapter Configuration -->
+  <MSTest>
+    <!-- Enable legacy mode, legacy setting are honored only when this flag is set to true -->
+    <ForcedLegacyMode>true</ForcedLegacyMode>
+  </MSTest>
+
+  <!-- Settings used for Ordered tests and MSTest based tests when running in legacy mode -->
+  <LegacySettings>
```
kaadhina: please check if we want to have descriptive section similar to datacollector

docs/configure.md
```diff
@@ -72,6 +73,8 @@ core principles for runsettings.
    test run via the test platform APIs. It may use it to read its own settings
    or take decisions (e.g. an adapter needs to disable appdomains if
    `DisableAppDomain` setting is provided via run settings).
+5. Users can now configure settings used for legacy scenarios like deployment,
+   scripts using run settings.
```
PBoraMSFT: Clarify that legacy scenarios are only applicable for MSTest v1 framework.

PBoraMSFT: Also add that this enables users to use all the settings that were available in testsettings using runsettings.

docs/configure.md
```diff
@@ -347,6 +366,59 @@ a test run.
...
+Here is an example:
+```xml
+<Runsettings>
+  <MSTest>
+    <!-- Enable legacy mode, legacy setting are honored only when this flag is set to true -->
+    <ForcedLegacyMode>true</ForcedLegacyMode>
+  </MSTest>
+
+  <LegacySettings>
+    <Deployment>
```
PBoraMSFT: Check case for Deployment

docs/configure.md
singhsarab: Covered these in the LegacySettings section below.

PR #176, opened 2019-02-26
Passing test sources parameter to data collectors

abhishkk:


AbhitejJohn:
@nohwnd : I'm not sure if this is currently supported today but if it is, can we merge and close?

nohwnd:
@AbhitejJohn  I will have to revisit this when I am done with all pri1 issues on 16.6 :)

PR #179, opened 2019-03-11
Runsettings provider extensibility

abhishkk:


RFCs/0026-RunSettings-Provider-Extensibility.md
```diff
@@ -0,0 +1,59 @@
...
+}
+```
+
+### ExtensionElement
+`ExtensionElement` is XmlElement of the target extension present in runsettings. If target extension is not present in runsettings, `null` value is passed for extensionElement argument.
+
+### RunSettingsProvider assembly naming convention 
+Separate assembly is not required for RunSettingsProvider. RunSettingsProvider can be specified in same assembly as of its target extension (like DataCollector, Logger). Thus, naming convention is similar to target extension (`*collector.dll` or `*logger.dll`)
+
+## Notes
```
singhsarab: What happens if there are multiple providers which register for the same node?

RFCs/0026-RunSettings-Provider-Extensibility.md
```diff
@@ -0,0 +1,59 @@
...
+1. Unlike DataCollectors/Loggers, RunSettingsProvider extension will always be plugged in test platform. RunSettingsProvider extensions are not required to be plugged in via runsettings or command line.
```
singhsarab: Also, it is not clear if the provider needs to be in the same dll as the extension or separate one.

RFCs/0026-RunSettings-Provider-Extensibility.md
```diff
@@ -0,0 +1,59 @@
...
+using Microsoft.VisualStudio.TestPlatform.ObjectModel;
+using Microsoft.VisualStudio.TestPlatform.ObjectModel.RunSettingsProvider;
+using Microsoft.VisualStudio.TestPlatform.ObjectModel.RunSettingsProvider.Attributes;
+
+[ExtensionUri("my://new/runsettingprovider/1.0")]
+[SettingsType("datacollector")]
+[SettingsUri("my://datacollector/1.0")]
+public class NewRunSettingsProvider : RunSettingsProvider
+{
+	public override XmlElement Process(SettingsProviderOptions settingsProviderOptions, XmlElement extensionElement)
```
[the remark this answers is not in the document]
abhishkk: Following are the suggestions:
1. ProcessExtensionElement
2. UpdateRunSettings
2. ProcessRunSettings
Which seems more appropriate?

RFCs/0026-RunSettings-Provider-Extensibility.md
abhishkk: All the providers will get the node one by one. So, if 1st provider updates the node, 2nd will get the updated node.

RFCs/0026-RunSettings-Provider-Extensibility.md
abhishkk: In line 54, titled `RunSettingsProvider assembly naming convention`, we have mentioned about it

PR #187, opened 2019-05-06
Adding doc for javascript testing using jstestadapter

karanjitsingh:
https://github.com/Microsoft/vstest-docs/pull/187/files?short_path=33a36a0#diff-33a36a0fdad4d4568892e07b0a212a8a


extensions/JSTestAdapter.md
```diff
@@ -0,0 +1,143 @@
+## Seamlessly running JavaScript tests with VSTest
```
singhsarab: There is already a folder inside docs called extensions, consider moving there if it makes sense to you.

karanjitsingh:
Pushing this in since prachi has reviewed offline

PR #196, opened 2019-07-01
[RFC] Merge dotnet test and vstest.

singhsarab:


RFCs/0027-Merge-Dotnet-Test-And-VsTest-Verbs.md
```diff
@@ -0,0 +1,46 @@
...
+
+## Issues with merging the two verbs
+
+### Irrelevant Arguments
+The arguments that are acceptable for the dotnet test for project/solution will not be applicable to the dotnet test when invoked for dlls.
+Here are possible ways to handle this:
+
+1. Ignore the arguments that are not valid.
+2. Throw a warning to the user, and continue with the execution.
+3. Throw an error, giving user an actionable message and stop the execution.
```
ViktorHofer: I think option 2 makes most sense. It should not fail the execution but ignoring the argument doesn't feel good either.

RFCs/0027-Merge-Dotnet-Test-And-VsTest-Verbs.md
```diff
@@ -0,0 +1,46 @@
...
+
+Note:
+1. For option (1), ignoring the arguments seems incorrect. If the user gives certain arguments, the expectation is that the arguments get honored. Ignoring the arguments without notify the user would be incorrect. If an argument like "--no-build" is not relevant 
+
+### Same arguments but different acceptable values
+There are arguments that are accepted by both dotnet test and dotnet vstest today. We will have make sure there is parity with respect to these arguments before merge.
+One such argument is "--framework". The values accepted by the dotnet test today are different from the ones accepted by dotnet vstest. Test platform today does not understand aliases. For example, ".NETCoreApp,Version=v1.0" is valid framework while netcoreapp1.0 is not valid framework for test platform.
+
+We will have to make sure all such arguments accept the super set of these values now. We might have to add another layer to convert the values into appropriate values that the platform understands.
```
ViktorHofer: ".NETCoreApp,Version=1.0" is the concatenation of `$(TargetFrameworkIdentifier),Version=$(TargetFrameworkVersion)`. The short form is `TargetFrameworkMoniker` (TFM), ie "netcoreapp1.0".

Why does dotnet vstest use the long form instead of the TFM as there is a 1:1 mapping between both? If there's no strong reason for that besides historical ones I would recommend to just use the TFM.

RFCs/0027-Merge-Dotnet-Test-And-VsTest-Verbs.md
```diff
@@ -0,0 +1,46 @@
...
+
+Should there be added intelligence to convert this argument ? or should this fall in invalid arguments ?
+
+## Possible Solution
+Parse the arguments given to "dotnet test" to detect dll/exe present in the arguments. If detected, call dotnet vstest.console.dll with the rest of the arguments.
+
+The above mentioned issues will need to be handled.
+
+Open questions:
+- What is the expected behavior if user gives both dlls and csproj ?
```
ViktorHofer: I think you shouldn't be able to specify both build artifacts and project files. I think throwing an error in the command line parser is appropriate here.

RFCs/0027-Merge-Dotnet-Test-And-VsTest-Verbs.md
singhsarab: I guess we should be using https://www.nuget.org/packages/NuGet.Frameworks/ then?
But for that, I guess we first need to move to netstandard2.0

ViktorHofer: @ericstj thoughts?

ericstj: Be careful in terminology here.  `$(TargetFrameworkMoniker)` is the full form and the one understood by the [FrameworkName](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.versioning.frameworkname?view=netframework-4.8) class.  For this there is a 1:1 mapping where a given framework can only be represented by one string.

`$(TargetFramework)` is the NuGet version of this, which doesn't have those characteristics.  Many strings can be used to represent the same framework.  So things like net45 and net4.5 are treated the same.

singhsarab: dotnet verbs today take -f or --framework as an argument, and this arguments supposedly is the NuGet version, as it accepts net451 as a valid value.

We on the other hand relied on FrameworkName, what I am thinking of doing is accept the input as NuGet version and then convert that into FrameworkName.
I thought NuGet.Framework has the mapping that I could leverage.

What do you guys suggest ?
The code sits here https://github.com/microsoft/vstest/blob/master/src/Microsoft.TestPlatform.ObjectModel/Framework.cs

ViktorHofer: I guess someone from the CLI team should be able to help here. 

cc @wli3 @livarcocc @nguerrera

singhsarab: @livarcocc @nguerrera Is that the right approach ?

wli3: cc @KathleenDollard 

I cannot recall any CLI command that is not default to short version. I don't think it will accept the longer version in most of the cases, since we usually just forward it to MSBuild as property TargetFramework

PR #202, opened 2019-08-08
 Specify Environment Variables in RunSettings RFC

vagisha-nidhi:


vagisha-nidhi:
@ViktorHofer Can you please go through this?

RFCs/0027-Specify-Environment-Variables-In-Runsettings.md
```diff
@@ -0,0 +1,32 @@
...
+      <DOTNET_ROOT>C:\ProgramFiles\dotnet</DOTNET_ROOT>
+      <SDK_PATH>C:\Codebase\Sdk</SDK_PATH>
+    </EnvironmentVariables>
+  </RunConfiguration>
+</RunSettings>
+
+
+```
+Since these environment variables should always be set when the test host is started, the tests should always run in a separate process.
+For this, the `/InIsolation` flag will be set when there are environment variables so that the test host is always invoked.
```
ViktorHofer: I think it makes sense to not set the InIsolation flag when env vars are set as others who aren't running in isolation want to use this feature as well, not for the testhost invocation but inside the test app.

RFCs/0027-Specify-Environment-Variables-In-Runsettings.md
```diff
@@ -0,0 +1,32 @@
...
+The different environment variables can be specified as element name and it's value.
+Below is a sample runsettings for passing environment variables.
+
+```csharp
+
+<?xml version="1.0" encoding="utf-8"?>
+<!-- File name extension must be .runsettings -->
+<RunSettings>
+  <RunConfiguration>
+    <EnvironmentVariables>
```
ViktorHofer: nit: do you usually use plural or singular? I.e. you have `RunSettings` (plural) but also `RunConfiguration` (singular).

RFCs/0027-Specify-Environment-Variables-In-Runsettings.md
vagisha-nidhi: In certain specific cases, we run the tests inside vstest.console process without starting the testhost process. Since we are planning to set the environment variables when starting the test host, these cases will not be encountered for. This flag will insure that the test host is launched with the correct variables.

RFCs/0027-Specify-Environment-Variables-In-Runsettings.md
vagisha-nidhi: For the `EnvironmentVariables`, it made sense to use plural to infer that multiple values are allowed.

RFCs/0027-Specify-Environment-Variables-In-Runsettings.md
ViktorHofer: Can we support both cases?
1) Set env vars before the testhost process is started
2) If tests are executed directly in vstest.console, set the env vars before invoking the tests.

ViktorHofer:
Should we close this PR?

PR #204, opened 2019-08-14
 Dotnet test on multi target project logging on multiple targets

vagisha-nidhi:
Issue related : https://github.com/microsoft/vstest/issues/1603


singhsarab:
@vagisha-nidhi Just conclude the RFC with the approach taken and why, and then merge it.

PR #210, opened 2019-09-26
Adding the documentation for Junit Logger

Siphonophora:
Hello, 

We just released a logger which produces Junit xml reports.


Siphonophora:
Hi, can this be merged? The logger is not brand new anymore if that was the original reason this didn't get attention.

Haplois:
Added, thank you.

PR #219, opened 2020-01-20
Passing test run paramters via cli

hvinett:


RFCs/0029-Passing-TestRunParameters-via-CLI.md
```diff
@@ -0,0 +1,50 @@
...
+## Summary
+Test run parameters given via cli are passed to runsettings.
+
+## Motivation
+Every time editing or creating one test run parameter in run settings at a time is a tedious task. So passing test run parameters via cli 
+removes burden of editing run settings every time.
+
+## Syntax
+
+`vstest.console.exe  abc.dll -- TestRunParaeters.Parameter(name="YourParameterName",value="YourParameterValue ")`
```
singhsarab: Nit: TestRunParameters*

Also, you might want to include BackSlash('\') I think that was part of the PR changes right ?

RFCs/0029-Passing-TestRunParameters-via-CLI.md
```diff
@@ -0,0 +1,50 @@
...
+
+<ul>
+<li>TestRunParameters as Node</li>
+<li>Parameter as child element of TestRunParamter node</li>
+<li>YourParamterName as Attribute name </li>
+<li>YourParameterValue as Attribute value </li>
+</ul>
+
+## Design
+Eventually the arguments are parsed and we get `attribute name` and `attribute value` of `test run parameter`.
```
singhsarab: Eventually I used when there is pretext. Since there is no pretext avoid using it.

New note for Design would look like:
"The arguments are parsed and `attribute name` and `attribute value` are fetched from the `TestRunParameter` argument.  Then, an xml element is overrode if exists with `attribute name` or created with `'Parameter'` as it's name and with attributes that are obtained by the above step. Finally, we get `TestRunParameter` node if exists or created and append the newly created xml element to list of child nodes."

RFCs/0029-Passing-TestRunParameters-via-CLI.md
hvinett: you mean this right?
TestRunParameters.Parameter(name=\"YourParameterName\",value=\"YourParameterValue\")`

PR #221, opened 2020-02-21
Add update to show support for vs 2019

vritant24:


docs/contribute.md
```diff
@@ -14,8 +14,9 @@ Clone the repository to a local directory. Rest of this article assumes
 ```
 
 If you're planning to use **Visual Studio** as development environment, please
-install `VS 2017 RC 2` and the `.NET Core and Docker (Preview)` workload installed. See download
-link [here](https://www.microsoft.com/net/core#windowsvs2017).
+install `VS 2019` and the `.NET Core and Docker (Preview)` workload installed. See download
+link [here](https://www.microsoft.com/net/core#windowsvs2017).
```
AbhitejJohn: @vritant24 : It looks like this download link is still pointing to vs2017. Can we update that please?

PR #227, opened 2020-03-27
Beef up docs for capturing dumps in CI

AArnott:
From several hours of effort to get this working on both Azure Pipelines and GitHub Actions, here are the doc updates I would recommend so other customers can do this with much less effort.

### DIFF
--- a/RFCs/0028-BlameCollector-Hang-Detection.md
+++ b/RFCs/0028-BlameCollector-Hang-Detection.md
@@ -1,44 +1,106 @@
 # 0028 Blame collector hang detection
 
 ## Summary
-Blame data collector now supports a new mode meant to help detect and fix hangs in test code. This mode does not introduce any perf hit on the test run as the proc dump process is only started after the specified timeout interval has elapsed. 
+
+Blame data collector now supports a new mode meant to help detect and fix hangs in test code. This mode does not introduce any perf hit on the test run as the proc dump process is only started after the specified timeout interval has elapsed.
 
 ## Motivation
+
 Whenever running tests in CI systems if a hang occurred it would generally lead to a timeout and abrupt cancelling of the CI pipeline without giving an oppurturnity for the test platform or other scripts to attach the necessary logs/dumps required to analyse the hang.
 
 ## Pre-Requisites
+
 This was introduced in testplatform version 16.4.0-preview-20191007-01 (This or a higher version is required)
 
 ## Working
+
 If the testhost does not send any messages to the datacollector for the specified duration then it is inferred as a hang and a dump is collected and the testhost process is killed to ensure any available attachments and logs are gracefully attached to the trx file and propagated up to the chain in case of a CI system for further analysis of the hang.
 
 ## Steps to configure
-Proc dump path needs to be set using the environment variable PROCDUMP_PATH
-Add the required settings as shown below in the runsettings file
 
-Sample runsettings
+Set the `PROCDUMP_PATH` environment variable to the full path to the directory containing
+the `procdump.exe` and `procdump64.exe` tools.
+This path may be `%ProgramData%\chocolatey\bin\` if you use `choco install procdump` to acquire these tools.
 
-```
+Add the required settings as shown below in the .runsettings file sample below:
+
+```xml
 <RunSettings>
-    <DataCollectionRunSettings>
-        <DataCollectors>
-            <DataCollector friendlyName="blame" enabled="True">
-                <Configuration>
-                    <CollectDumpOnTestSessionHang TestTimeout="300000" DumpType="mini" />
-                    <ResultsDirectory>%AGENT_TEMPDIRECTORY%\TestResults</ResultsDirectory>
-                </Configuration>
-            </DataCollector>
-        </DataCollectors>
-    </DataCollectionRunSettings>
+  <DataCollectionRunSettings>
+    <DataCollectors>
+      <DataCollector friendlyName="blame" enabled="True">
+        <Configuration>
+          <CollectDumpOnTestSessionHang TestTimeout="300000" DumpType="mini" />
+          <ResultsDirectory>%AGENT_TEMPDIRECTORY%</ResultsDirectory>
+        </Configuration>
+      </DataCollector>
+    </DataCollectors>
+  </DataCollectionRunSettings>
 </RunSettings>
 ```
 
-Note: This works along with (but can also be used independent of) the existing collect dump option introduced in https://github.com/microsoft/vstest-docs/blob/master/RFCs/0024-Blame-Collector-Options.md
+Note: This works along with (but can also be used independent of) the existing collect dump option introduced in [RFC 0024](0024-Blame-Collector-Options.md).
 
 ## Supported options
-DumpType: If you choose to collect a full process dump. It takes values mini/full. By default, a mini dump will be created.
-TestTimeout: Duration of inactivity in milliseconds (no test events from the test host) after which the data collector assumes a hang has occurred and proceeds to collect a dump and kill the test host process.
-ResultsDirectory: Location to temporarily store the collected process dump
 
-## VsTest task usage
+|Option|Description|
+|--|--|
+`DumpType`|If you choose to collect a full process dump. It takes values `mini`/`full`. By default, a mini dump will be created.
+`TestTimeout`|Duration of inactivity in milliseconds (no test events from the test host) after which the data collector assumes a hang has occurred and proceeds to collect a dump and kill the test host process.
+`ResultsDirectory`|The path must exist, but files will not be permanently stored there. Test results ultimately go to a `TestResults` folder under the test project directory by default, or where the `--results-directory` parameter specifies.
+
+## Invoking the tests
+
+Add the `--settings path-to\test.runsettings` parameter to `dotnet test` or `vstest.console.exe`.
+Dump collection with Blame data collector is only supported on Windows for now. For non-Windows agents, omit the .runsettings file or use one that does not activate the Blame collector.
+
+Use the `--results-directory` to control where the `Sequence_<guid>.xml` and `.dmp` files are placed.
+
+## Outputs
+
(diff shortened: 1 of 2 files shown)


RFCs/0028-BlameCollector-Hang-Detection.md
```diff
@@ -1,44 +1,88 @@
...
 ## Supported options
-DumpType: If you choose to collect a full process dump. It takes values mini/full. By default, a mini dump will be created.
-TestTimeout: Duration of inactivity in milliseconds (no test events from the test host) after which the data collector assumes a hang has occurred and proceeds to collect a dump and kill the test host process.
-ResultsDirectory: Location to temporarily store the collected process dump
+
+|Option|Description|
+|--|--|
+`DumpType`|If you choose to collect a full process dump. It takes values `mini`/`full`. By default, a mini dump will be created.
+`TestTimeout`|Duration of inactivity in milliseconds (no test events from the test host) after which the data collector assumes a hang has occurred and proceeds to collect a dump and kill the test host process.
+`ResultsDirectory`|The path must exist, but it is not used. Test results go to a `TestResults` folder under the test project directory when `dotnet test` is used locall, or under `$(Agent.TempDirectory)\<guid>` when the `DotNetCoreCLI` task is used on Azure Pipelines.
```
ShreyasRmsft: locally

ShreyasRmsft: This path is used to temporarily store the dump file but eventually the testplatform copies the dumpfile (and all attachments in general) into relative folders next to the trx.

ShreyasRmsft: can we say vstest.console.exe/dotnet test

ShreyasRmsft: Let's move this to a separate section at the botttom pertaining to usage in Azure Pipelines

RFCs/0028-BlameCollector-Hang-Detection.md
```diff
@@ -1,44 +1,88 @@
...
+
+## Invoking the tests
+
+Add the `--settings path-to\test.runsettings` parameter to `dotnet test` or `vstest.console.exe`.
+Note that this will cause non-Windows agents to fail, so only add this switch on Windows agents.
```
ShreyasRmsft: Can we please rephrase this to 'Dump collection with Blame data collector is only supported on Windows for now.'

RFCs/0028-BlameCollector-Hang-Detection.md
```diff
@@ -1,44 +1,88 @@
...
+
+## Outputs
+
+The `Sequence_guid.xml` and .dmp files will be placed under the test project source directory in a `TestResults\<guid>\` folder.
```
ShreyasRmsft: This is only created if the blame logger is enabled. Configuring hang dump alone does not create this. This is already referenced in the previous Blame RFC, maybe we can add a line under a section called Recommended Usage to say use this along with the blame logger for best results.

  <Loggers>
      <Logger friendlyName="blame" enabled="True" />
    </Loggers>

RFCs/0028-BlameCollector-Hang-Detection.md
```diff
@@ -1,44 +1,88 @@
...
+
+The `DotNetCoreCLI` task will try to upload these files as attachments, but any dump file exceeding 75MB will fail (and full dumps are almost surely going to exceed this limit).
```
ShreyasRmsft: This limitation is only with the dotnet task, won't be a problem with the vstest task.

I'd prefer if we actually make 2 separate sections one for vstest and one for dotnet test to make sure the limitations are called out accordingly

RFCs/0028-BlameCollector-Hang-Detection.md
AArnott: good catch.

AArnott: I've updated the wording, avoiding the reference to trx because by default that file isn't created.

AArnott: Done

RFCs/0028-BlameCollector-Hang-Detection.md
AArnott: Sure. I added one more sentence though to retain the point that would have saved me some time:

> Dump collection with Blame data collector is only supported on Windows for now. For non-Windows agents, omit the .runsettings file or use one that does not activate the Blame collector.

Without this extra sentence it is not clear that the user has to *avoid* asking for the blame collector since it might just 'silently no-op' on other platforms.

RFCs/0028-BlameCollector-Hang-Detection.md
AArnott: Isn't the blame logger required to be enabled before turning on the dmp collector? To turn on dumping you need xml elements inside another blame element:

```xml
      <DataCollector friendlyName="blame" enabled="True">
        <Configuration>
          <CollectDump DumpType="full" />
          <CollectDumpOnTestSessionHang TestTimeout="30000" DumpType="full" />
          <!-- This ResultsDirectory must exist, but it is totally ignored. -->
          <ResultsDirectory>%BUILD_ARTIFACTSTAGINGDIRECTORY%</ResultsDirectory>
        </Configuration>
      </DataCollector>
```

What does this look like without the sequence file but with dmp files?

RFCs/0028-BlameCollector-Hang-Detection.md
AArnott: Sounds good.

RFCs/0028-BlameCollector-Hang-Detection.md
ShreyasRmsft: Actually this is the section responsible for activating the blame logger (separate entity actually) which generates the sequence file

```
  <LoggerRunSettings>
        <Loggers>	    <Loggers>
            <Logger friendlyName="blame" enabled="True" />	      <Logger friendlyName="blame" enabled="True" />
        </Loggers>	    </Loggers>
    </LoggerRunSettings>	  </LoggerRunSettings> 
```

ShreyasRmsft: The logger and the blame datacollector can be used independent of each other. Infact they leverage 2 different extensibility points of vstest

AArnott: I observe the Sequence file being generated with or without the `Logger` snippet you just prescribed @ShreyasRmsft. As long as this block is present, the Sequence file is created on crashes:

```xml
  <DataCollectionRunSettings>
    <DataCollectors>
      <DataCollector friendlyName="blame" enabled="True">
        <Configuration>
          <CollectDump DumpType="full" />
          <CollectDumpOnTestSessionHang TestTimeout="30000" DumpType="full" />
          <!-- This ResultsDirectory must exist, but it is totally ignored. -->
          <ResultsDirectory>d:\temp</ResultsDirectory>
        </Configuration>
      </DataCollector>
    </DataCollectors>
  </DataCollectionRunSettings>
```

ShreyasRmsft: Are you adding the --blame option via the console?

RFCs/0028-BlameCollector-Hang-Detection.md
```diff
@@ -1,44 +1,106 @@
...
+Use the `--results-directory` to control where the `Sequence_<guid>.xml` and `.dmp` files are placed.
+
+## Outputs
+
+The `Sequence_guid.xml` and .dmp files will be placed under the test project source directory in a `TestResults\<guid>\` folder.
+
+Note that the test runner may report a failure to produce the dump file, such as the one shown below:
+
+```
+Data collector 'Blame' message: System.IO.FileNotFoundException: Collect dump was enabled but no dump file was generated.
```
ShreyasRmsft: Also this is probably coming from the OG blame and not the hang detection part. I took a look at the runsettings you are using and it has the `<CollectDump DumpType="full" />` node which also configures crash dumps.

RFCs/0028-BlameCollector-Hang-Detection.md
AArnott: No. I only am adding the `-s` switch (and sometimes the `-r` switch).

ShreyasRmsft: np, going ahead and merging this. I had a few things i wanted to enhance on the vstest usage side, will do it in a separate pr.

RFCs/0028-BlameCollector-Hang-Detection.md
AArnott: Would you like me to remove all traces of crash dump tips? I could potentially move them to the other RFC. 
IMO though, anyone who goes to the trouble to configure dump collection at all *probably* wants both hang and crash dumps. So combining the two into one doc with a few notes to say "you can turn this individual feature on/off like this..." makes the most sense.

ShreyasRmsft: I would prefer if we did this. Maybe in a separate pr.

The reason being crash detection mode has a perf hit associated with it that scales linearly with the number of exceptions being thrown by the test code. Proc dump has a first chance hook on these and even if they are handled exceptions there's a perf hit.

AArnott: ProcDump has managed hooks? I had no idea. A quick review of the options confirms what you say, but I'm disappointed if the perf hit of first chance exception monitoring applies even when you *don't* pass in the switches to enable that behavior.

ShreyasRmsft: Sadly this is true, in fact when this was initially written folks didn't realize this and the crash detection was enabled by default. Once it got enabled for all customers we got reports that there is a sudden perf hit. 

I investigated the cause and realized this was it. And that the switch that says "include first chance exceptions" just indicates whether to write the dump to disk when it occurs not whether it should monitor first chance exceptions or not. But it made sense because that's the only way proc dump would be able to capture the state of the process as is.

If it had relied on unhandledException hook then the stack would have unwound by the time the proc dump got notified.

PR #235, opened 2020-06-24
RFC for multi test run finalization

jakubch1:


RFCs/0031-Multi-Test-Run-Finalization.md
```diff
@@ -0,0 +1,137 @@
+# 0031 Multi Test Run Finalization
+
+# Summary
+Introduce APIs to perform data attachments reprocessing (e.g. combining, merging) after two or more independent test executions.
```
AbhitejJohn, line 4: Perhaps we could highlight the scenario a little better. 
Suggestion: "This document details a data collector extensibility point to reporcess (combine/merge) attachments obtained across test executions in both design mode and commandline scenarios"

I'm assuming we refer to the non-commandline mode as "design mode". Also please consider using `data collector attachments` instead.

RFCs/0031-Multi-Test-Run-Finalization.md
```diff
@@ -0,0 +1,137 @@
...
+
+# Motivation
+Today when Test Platform executes tests in parallel only code coverage reports are merged (attachments with uri: `datacollector://microsoft/CodeCoverage/2.0`). For other attachments processing is skipped and all attachments are returned by Test Platform.
+
+The [dotnet test](https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-test) command is used to execute unit tests in a given solution. The `dotnet test` command builds the solution and runs a test host application for each test project in the solution. However, currently there is no way to combine/merge attachments associated with each project execution. Coverage reports are not merged.
+
+When `Run All Tests` is performed in VS, tests for projects can be executed separately. In this case also combining/merging of data attachments is not performed. Coverage reports are not merged. `Analyze Code Coverage for All Tests` is showing coverage report for only one of the test projects.
```
AbhitejJohn, line 11: `can be executed separately` -> `can be executed separately based on target platform/target framework amongst other criteria`

AbhitejJohn, line 11: `is showing coverage report for only one of the test projects` -> `is showing coverage report for only some test projects in the run`

RFCs/0031-Multi-Test-Run-Finalization.md
```diff
@@ -0,0 +1,137 @@
...
+
+# Proposed Changes
+1. Introduce a `IDataCollectorAttachments` interface which can be implemented by Test Platform extensions and provide custom logic to combine/merge attachments. Below interface should be used only for providing logic to combine/merge information from data attachments from independent test executions. This interface should **not be** used for modifying separated data attachment. Test Platform will invoke `HandleDataCollectionAttachmentSets` only if at least 2 data attachments related to Collector (through `GetExtensionUris`) are created by independent test executions.
```
AbhitejJohn, line 14: `This interface should **not be** used for modifying separated data attachment` -> Can you elaborate on this please?

AbhitejJohn: Also please consider using data collector attachment as a term throughout this document.

RFCs/0031-Multi-Test-Run-Finalization.md
```diff
@@ -0,0 +1,137 @@
...
+
+```
+namespace Microsoft.VisualStudio.TestPlatform.ObjectModel.DataCollection
+{
+    /// <summary>
+    /// Interface for data collectors add-ins that choose to handle generated attachments
+    /// </summary>
+    public interface IDataCollectorAttachments
```
AbhitejJohn, line 22: Could we call this an `IDataCollectorAttachmentProcessor` instead? I'd expect metadata information for attachments with `IDataCollectorAttachments`

RFCs/0031-Multi-Test-Run-Finalization.md
```diff
@@ -0,0 +1,137 @@
...
+        /// <param name="cancellationToken">Cancellation token</param>
+        /// <returns>Attachments after reprocessing</returns>
+        ICollection<AttachmentSet> HandleDataCollectionAttachmentSets(ICollection<AttachmentSet> attachments, IProgress<int> progressReporter, IMessageLogger logger, CancellationToken cancellationToken);
+    }
+}
+```
+
+Method `GetExtensionUris` should provide all Uris for data attachments which are handled by current collector. Test platform will provide to collector only attachments with such Uris. It's also required that result of method `HandleDataCollectionAttachmentSets` will contain only attachments with such Uris.
+
+`IsAssociative` should indicate if `HandleDataCollectionAttachmentSets` is [associative](https://en.wikipedia.org/wiki/Associative_property).
```
AbhitejJohn, line 49: I like the idea behind this. However do you think this should be the testplatforms responsibility to check isAssociative for each processor and handle to logic to call collectors accordingly? How about if the test platform passes in a context to the processors about whether this is the last test run in a series or something in between? Something like a DataCollectionProcessorContext passed into HandleDataCollectionAttachments which would contain a `IsLastBatch` bool. 
I'd recommend also sending through runsettings in the context, so processors can choose to operate off of some user provider settings.

RFCs/0031-Multi-Test-Run-Finalization.md
```diff
@@ -0,0 +1,137 @@
...
+
+        /// <summary>
+        /// Dispatch MultiTestRunFinalizationProgress event to listeners.
+        /// </summary>
+        /// <param name="finalizationProgressEventArgs">Finalization Progress event args.</param>
+        void HandleMultiTestRunFinalizationProgress(MultiTestRunFinalizationProgressEventArgs finalizationProgressEventArgs);
+    }
+}
+```
+Interface provides callbacks from Multi Test Run Finalization process. For every finalization process `HandleMultiTestRunFinalizationComplete` will be called once and will provide last chunk or all data attachments. During finalization process `HandleFinalisedAttachments` can be invoked several times providing data attachments that are already processed. Method `HandleMultiTestRunFinalizationProgress` will be invoked every time when `progressReporter` is used by `IDataCollectorAttachments` implementation and will provide information about current collector: progress, uris and index of collector. Additionally event will contain also number of collectors.
```
AbhitejJohn, line 109: `For every finalization process` -> What does a finalization process mean?

AbhitejJohn, line 109: Can we document what "MultiTestRunFinalizationProgressEventArgs" contains too?

RFCs/0031-Multi-Test-Run-Finalization.md
```diff
@@ -0,0 +1,137 @@
...
+```
+/// <summary>
+/// Provides back all attachments to TestPlatform for additional processing (for example merging)
+/// </summary>
+/// <param name="attachments">Collection of attachments</param>
+/// <param name="multiTestRunCompleted">Indicates that all test executions are done and all data is provided</param>
+/// <param name="collectMetrics">Enables metrics collection</param>
+/// <param name="multiTestRunFinalizationCompleteEventsHandler">EventHandler to receive session complete event</param>
+/// <param name="cancellationToken">Cancellation token</param>        
+Task FinalizeMultiTestRunAsync(IEnumerable<AttachmentSet> attachments, bool multiTestRunCompleted, bool collectMetrics, IMultiTestRunFinalizationEventsHandler eventsHandler, CancellationToken cancellationToken);
```
AbhitejJohn, line 74: We should also mention the raw json messages passed through to vstest.console so any clients talking through json (and not using the managed translation layer) can benefit from this as well.

RFCs/0031-Multi-Test-Run-Finalization.md
```diff
@@ -0,0 +1,137 @@
...
+
+4. Use above logic to combine/merge data attachments for parallel test executions and VS scenarios (e.g. `Run All Tests`, `Analyze Code Coverage for All Tests`). In case of `Analyze Code Coverage for All Tests` VS will use `vstest.console` in Design Mode and merge all code coverage reports. VS will show full code coverage report for all test projects.
+
+5. When [dotnet test](https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-test) command is used to execute unit tests in a given solution, new console app `Orchestrator` will be executed.
+  * `Orchestrator` for every project inside solution will start `dotnet test` command, using `Process` with output redirected. Output for every project will be printed to stdout with some mutex to not mix output from children. 
+  * Additionally new test platform extension will be introduced and will propagate data about:
+    - Attachments
+    - Test results statistics
+  back to `Orchestrator`.
+  * In parallel with test executions `Orchestrator` will start `vstest.console` in Design Mode.
```
AbhitejJohn, line 120: `in Design mode` -> `in a variation of design mode`

RFCs/0031-Multi-Test-Run-Finalization.md
```diff
@@ -0,0 +1,137 @@
...
+  * Whenever at least 2 test executions are finished `Orchestrator` will invoke `FinalizeMultiTestRunAsync` and provide all attachments from those test executions that finished. Parameter `multiTestRunCompleted` will be set to `false`. Test platform will provide data attachments only to associative collectors.
+  * When all test exections are done `Orchestrator` will provide all attachments back through `FinalizeMultiTestRunAsync` with `multiTestRunCompleted` set to `true`. Test Platfrom will use all available collectors to process data attachments.
```
AbhitejJohn, line 122: `Platfrom` -> `Platform`

RFCs/0031-Multi-Test-Run-Finalization.md
```diff
@@ -0,0 +1,137 @@
...
+  * When all attachments are merged `Orchestrator` will display information about data attachments to standard output. 
+  * Finally `Orchestator` will combine all tests results print it to standard output.
```
AbhitejJohn, line 124: Did you mean overall test run statistics here?

RFCs/0031-Multi-Test-Run-Finalization.md
```diff
@@ -0,0 +1,137 @@
...
+
+
+4. Use above logic to combine/merge data attachments for parallel test executions and VS scenarios (e.g. `Run All Tests`, `Analyze Code Coverage for All Tests`). In case of `Analyze Code Coverage for All Tests` VS will use `vstest.console` in Design Mode and merge all code coverage reports. VS will show full code coverage report for all test projects.
+
+5. When [dotnet test](https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-test) command is used to execute unit tests in a given solution, new console app `Orchestrator` will be executed.
+  * `Orchestrator` for every project inside solution will start `dotnet test` command, using `Process` with output redirected. Output for every project will be printed to stdout with some mutex to not mix output from children.
```
AbhitejJohn, line 115: a block/flow diagram would be very useful for this.

RFCs/0031-Multi-Test-Run-Finalization.md
AbhitejJohn: I'm assuming we wouldn't be sending back test case and test result objects in this mode. Do we know what this interface will look like?

RFCs/0031-Multi-Test-Run-Finalization.md
jakubch1: fixed

RFCs/0031-Multi-Test-Run-Finalization.md
jakubch1: I totally agree. I just didn't want to modify existing interface. Can we keep existing interface and mark it as deprecated? After my changes it will not be used anywhere. Additionally we do not search for this interface in extensions.

RFCs/0031-Multi-Test-Run-Finalization.md
jakubch1: fixed

RFCs/0031-Multi-Test-Run-Finalization.md
jakubch1: Added explanation inside section 2 about method FinalizeMultiTestRunAsync

RFCs/0031-Multi-Test-Run-Finalization.md
jakubch1: Added more info and used data collector attachment in all places. Let's still discuss this.

RFCs/0031-Multi-Test-Run-Finalization.md
jakubch1: Having isAssociative on processor interface gives us possibility to create performance improvements which can be required in the future. For example TestPlatform can keep map of uri -> isAssociative for all processors. Such map can be provided back to VsTestConsoleWrapper through messages. Wrapper can perform filtering immediately of all incoming attachments.

Not having such flag will cause that we always need to provide all attachments to all processors. For very large project with hundreds of attachments even with lastBatch false all attachments needs to be propagated to all processors.

RFCs/0031-Multi-Test-Run-Finalization.md
jakubch1: added

RFCs/0031-Multi-Test-Run-Finalization.md
jakubch1: Documented those messages, but do we really want to have such data inside RFC? How we manage consistency? Can we replace it with links to code when we merge this to master?

RFCs/0031-Multi-Test-Run-Finalization.md
jakubch1: yeap, i have to be more carefull :)

RFCs/0031-Multi-Test-Run-Finalization.md
jakubch1: I would create new interface for this. It could be similar to:

interface IVsTestConsoleListener
{
     public Task ListenForTestRunEventsAsync(ITestRunEventsHandler testRunEventsHandler);
}

another possibility that maybe we can execute for dotnet everything in design mode. In both cases its not changing existing Multi Test Run Finalization design. I would not include it in this document. WDYT?

RFCs/0031-Multi-Test-Run-Finalization.md
jakubch1: diagram added

RFCs/0031-Multi-Test-Run-Finalization.md
jakubch1: I think this whole finalization should be used to combine data from 2 or more attachments. Any other logic on attachment can be done before finalization starts, for example on TestRunComplete event, or even during initial creation of attachment.

RFCs/0031-Multi-Test-Run-Finalization.md
```diff
@@ -0,0 +1,304 @@
...
+    - Attachments
+    - Test results statistics
+  back to `Orchestrator`.
+  * In parallel with test executions `Orchestrator` will start `vstest.console` in Design Mode.
+  * Whenever at least 2 test executions are finished `Orchestrator` will invoke `FinalizeMultiTestRunAsync` and provide all attachments from those test executions that finished. Parameter `multiTestRunCompleted` will be set to `false`. Test platform will provide data collector attachments only to associative processors.
+  * When all test exections are done `Orchestrator` will provide all attachments back through `FinalizeMultiTestRunAsync` with `multiTestRunCompleted` set to `true`. Test Platform will use all available processors to process data collector attachments.
+  * When all attachments are merged `Orchestrator` will display information about data collector attachments to standard output. 
+  * Finally `Orchestator` will combine all tests results statistics and print it to standard output.
+
+![dotnet-test finalization](Images/finalization.png)
```
AbhitejJohn, line 128: >dotnet-test finalization](Images/finalization.png [](start = 2, length = 49)

Thanks, this makes things a lot more clearer. Few comments on the image:
1. Aren't the ports for each dotnet test process different? 
2. It looks like the communication between vstest.console and the orchestrator need to be updated. Should the arrows point in the reverse direction? The finalization also happens in the orchestrator doesn't it?
3. Can we indicate that intermediate test results are still sent back to the parent host?
4. When we say vstest.console, I'm assuming we are referring to the parent vstest/test task that would spawn off dotnet vstest/test with the orchestrator. We could specify it as so.

RFCs/0031-Multi-Test-Run-Finalization.md
```diff
@@ -0,0 +1,304 @@
+# 0031 Multi Test Run Finalization
+
+# Summary
+This document details a data collector extensibility point to reporcess (combine/merge) attachments obtained across test executions in both design mode and commandline scenarios.
```
AbhitejJohn, line 4: >reporcess [](start = 62, length = 9)

nit: reprocess.

RFCs/0031-Multi-Test-Run-Finalization.md
AbhitejJohn: I agree with deprecating the old interface because there was no chance this ever worked for the intended workflow and putting out a new one with the name change.

---
In reply to: [446883700](https://github.com/microsoft/vstest-docs/pull/235#discussion_r446883700) [](ancestors = 446883700)

RFCs/0031-Multi-Test-Run-Finalization.md
jakubch1: The main process is second thing (Orchestrator). It made it not clear. I will fix it

RFCs/0031-Multi-Test-Run-Finalization.md
```diff
@@ -0,0 +1,304 @@
...
+
+When `Run All Tests` is performed in VS, tests for projects can be executed separately based on target platform/target framework amongst other criteria. In this case also combining/merging of data collector attachments is not performed (e.g. code overage reports are not merged). `Analyze Code Coverage for All Tests` is showing coverage report for only some test projects in the run.
+
+# Proposed Changes
+
+Introduce a new `IDataCollectorAttachmentProcessor` interface which can be implemented by Test Platform extensions and provide custom logic to reporcess (combine/merge) data collector attachments. Below interface should be used only for providing logic to reprocess information from data collector attachments from independent test executions. In other words data collector attachment should be modified/created using data from other data collector attachments. This interface should **not be** used for modifying any single data collector attachment. Test Platform will invoke `ProcessAttachmentSets` only if at least 2 data collector attachments related to processor (through `GetExtensionUris`) are created by independent test executions.
```
AbhitejJohn, line 15: >IDataCollectorAttachmentProcessor [](start = 17, length = 33)

We'd also want to add how these processors would be discovered. Would it be through runsettings? Would they need to be copied over to the output directory? Would they need to be placed in the Extensions folder? I'd prefer if we do not require these extensions to ship with TestPlatform.

RFCs/0031-Multi-Test-Run-Finalization.md
AbhitejJohn: How about renaming it as `supportsIncrementalProcessing` in that case?

---
In reply to: [446938059](https://github.com/microsoft/vstest-docs/pull/235#discussion_r446938059) [](ancestors = 446938059)

AbhitejJohn: I was also thinking of an event handler mechanism like we have for DataCollector interface where processors can subscribe to intermediate attachment processing events or a final attachment processing event and all the TestPlatform would need to do is to invoke the right events. That would make it coherent with the model we have but that ordering of data collectors wouldn't be natural in that model. For ordering I'd expect data collectors to add an attribute specifying a number (with a lower number being the highest order) and the TestPlatform can invoke adapters in that order.

---
In reply to: [447192771](https://github.com/microsoft/vstest-docs/pull/235#discussion_r447192771) [](ancestors = 447192771,446938059)

RFCs/0031-Multi-Test-Run-Finalization.md
jakubch1: @AbhitejJohn  I don't have any experience with this. What do you recommend?

RFCs/0031-Multi-Test-Run-Finalization.md
AbhitejJohn: Pointing to source code with a fixed SHA so it doesn't change sounds good. For the protocol, I was referring to a something similar to this: [https://github.com/microsoft/vstest-docs/blob/master/RFCs/0012-Editors-API-RevisionUpdate-15.4.md](https://github.com/microsoft/vstest-docs/blob/master/RFCs/0012-Editors-API-RevisionUpdate-15.4.md). We could put that out into a separate doc to make it more discoverable.

---
In reply to: [446987820](https://github.com/microsoft/vstest-docs/pull/235#discussion_r446987820) [](ancestors = 446987820)

RFCs/0031-Multi-Test-Run-Finalization.md
jakubch1: fixed

RFCs/0031-Multi-Test-Run-Finalization.md
jakubch1: Do we need intermediate test results? I was thinking of taking all output from inner process and print it to stdout.

RFCs/0031-Multi-Test-Run-Finalization.md
AbhitejJohn: Yes, this document is primarily focused on flushing out the IDE scenario. The CLI workflows I'd expect to be more detailed in another doc based on the overview you've detailed here, when we get to that work. As long as we've thought about how these specific data collector attachment processing interfaces would apply to CLI workflows I think this document would do justice and I think we've achieved that. Perhaps that's something you'd want to add to the summary of this doc.

---
In reply to: [447004623](https://github.com/microsoft/vstest-docs/pull/235#discussion_r447004623) [](ancestors = 447004623)

RFCs/0031-Multi-Test-Run-Finalization.md
AbhitejJohn: I meant some indication that stdout would still be populated with test results in real time. We do not need individual test cases/results specially for this.

---
In reply to: [447198580](https://github.com/microsoft/vstest-docs/pull/235#discussion_r447198580) [](ancestors = 447198580)

RFCs/0031-Multi-Test-Run-Finalization.md
```diff
@@ -0,0 +1,304 @@
...
+        public bool IsCanceled { get; private set; }
+
+        /// <summary>
+        /// Gets the error encountered during the finalization of the test runs. Null if there is no error.
+        /// </summary>
+        [DataMember]
+        public Exception Error { get; private set; }
+
+        /// <summary>
+        /// Get or Sets the Metrics
```
AbhitejJohn, line 176: >Metrics [](start = 28, length = 7)

Nit: mention that this is for telemetry.

RFCs/0031-Multi-Test-Run-Finalization.md
```diff
@@ -0,0 +1,304 @@
...
+
+3. Class `MultiTestRunFinalizationPayload` defines message payload connected to message ID `"MultiTestRunFinalization.Start"`
+
+```
+namespace Microsoft.VisualStudio.TestPlatform.ObjectModel.Client
+{
+    /// <summary>
+    /// Class used to define the MultiTestRunFinalizationPayload sent by the Vstest.console translation layers into design mode
+    /// </summary>
+    public class MultiTestRunFinalizationPayload
```
AbhitejJohn, line 242: >MultiTestRunFinalizationPayload [](start = 17, length = 31)

We probably could defer this to the editor API specification doc I was referring to earlier. That can happen in a separate PR as well if you'd prefer so you wouldn't not to block any code changes based on that.

RFCs/0031-Multi-Test-Run-Finalization.md
```diff
@@ -0,0 +1,304 @@
...
+
+4. Class `MultiTestRunFinalizationCompletePayload` defines message payload connected to message ID `"MultiTestRunFinalization.Complete"`
+
+```
+namespace Microsoft.VisualStudio.TestPlatform.CommunicationUtilities.ObjectModel
+{
+    /// <summary>
+    /// Multi test run finalization complete payload.
+    /// </summary>
+    public class MultiTestRunFinalizationCompletePayload
```
AbhitejJohn, line 273: >MultiTestRunFinalizationCompletePayload [](start = 17, length = 39)

This and the class below can also fit into the editor API spec above.

AbhitejJohn: The level of detail in the doc could be put in the format that earlier spec docs had.

---
In reply to: [447204380](https://github.com/microsoft/vstest-docs/pull/235#discussion_r447204380) [](ancestors = 447204380)

RFCs/0031-Multi-Test-Run-Finalization.md
AbhitejJohn: It wouldn't be unreasonable, for this effort, to make this similar to our data collector extension discovery - which allows extension discovery through a testadapterpath specification in runsettings/commandline or through the extensions folder. As part of the performance efforts, we have another work item to optimize sending in the relevant extensions to the platform which should cover better nuget package support as well. 

---
In reply to: [447196449](https://github.com/microsoft/vstest-docs/pull/235#discussion_r447196449) [](ancestors = 447196449)

RFCs/0031-Multi-Test-Run-Finalization.md
```diff
@@ -0,0 +1,304 @@
...
+    {
+        /// <summary>
+        /// Gets the attachments Uris, which are handled by processor
+        /// </summary>
+        IEnumerable<Uri> GetExtensionUris();
+
+        /// <summary>
+        /// Indicates whether ProcessAttachmentSets is associative (e.g. f([a, b, c]) = f([f([a, b]), c]) = f([a, f([b, c])]))
+        /// </summary>
+        bool IsAssociative  { get; }
```
drognanar, line 33: should we specifcally use associativity, or perhaps we should just allow for `SupportsPartialMerge`, where for instance `f([a, b, c]) == f(f([a, c]), b)` would also hold.

also, what's the lifetime of these?

perhaps we could have a streaming API where we can pass in ProcessAttachments(...) and then have a MergeAttachments(...) method that does the merge. If a processor can process the attachments on the fly, it would.

RFCs/0031-Multi-Test-Run-Finalization.md
```diff
@@ -0,0 +1,304 @@
...
+
+        /// <summary>
+        /// Reprocess attachments generated by independent test executions
+        /// </summary>
+        /// <param name="attachments">Attachments to be processed</param>
```
drognanar, line 38: if multiple GetExtensionUris are specified and a collection is associative, will we ensure that we wouldn't give a subset of a attachment set from a run? for instance it I'd get attachments `[pdb1, coverage1, pdb2, coverage2]` then we'd never call `ProcessAttachmentsSet([pdb1, coverage1, pdb2])`?

RFCs/0031-Multi-Test-Run-Finalization.md
```diff
@@ -0,0 +1,304 @@
...
+```
+/// <summary>
+/// Provides back all attachments to TestPlatform for additional processing (for example merging)
+/// </summary>
+/// <param name="attachments">Collection of attachments</param>
+/// <param name="multiTestRunCompleted">Indicates that all test executions are done and all data is provided</param>
+/// <param name="collectMetrics">Enables metrics collection</param>
+/// <param name="multiTestRunFinalizationCompleteEventsHandler">EventHandler to receive session complete event</param>
+/// <param name="cancellationToken">Cancellation token</param>        
+Task FinalizeMultiTestRunAsync(IEnumerable<AttachmentSet> attachments, bool multiTestRunCompleted, bool collectMetrics, IMultiTestRunFinalizationEventsHandler eventsHandler, CancellationToken cancellationToken);
```
drognanar, line 75: what kind of metrics are being collected here? as a caller I'm not sure if I know what value to pass into this API.
perhaps this should be placed under a class or there should be default value provided?

it seems that test run completion also provides metrics. are those also mergeable?

RFCs/0031-Multi-Test-Run-Finalization.md
```diff
@@ -0,0 +1,304 @@
...
+
+By default `IsAssociative` should be `False`, unless processing can take longer time and it's beneficial to start the process as soon as possible.
+
+
+
+2. Introduce a new `FinalizeMultiTestRunAsync` method in [IVsTestConsoleWrapper](https://github.com/microsoft/vstest/blob/master/src/Microsoft.TestPlatform.VsTestConsole.TranslationLayer/Interfaces/IVsTestConsoleWrapper.cs) interface:
+
+```
+/// <summary>
+/// Provides back all attachments to TestPlatform for additional processing (for example merging)
```
drognanar, line 68: it would be nice to write somewhere the example of usage. that when you RunTests with a ITestRunEventsHandler, then the attachments coming from multiple HandleTestRunComplete invocations can be merged.

it took me a while for this method can be used

RFCs/0031-Multi-Test-Run-Finalization.md
```diff
@@ -0,0 +1,304 @@
...
+    {
+        /// <summary>
+        /// Dispatch MultiTestRunFinalizationComplete event to listeners.
+        /// </summary>
+        /// <param name="finalizationCompleteEventArgs">Finalization Complete event args.</param>
+        /// <param name="attachments">Last set of processed attachment sets.</param>
+        void HandleMultiTestRunFinalizationComplete(MultiTestRunFinalizationCompleteEventArgs finalizationCompleteEventArgs, IEnumerable<AttachmentSet> lastChunk);
+
+        /// <summary>
+        /// Dispatch FinalisedAttachments event to listeners.
```
drognanar, line 98: nit: should this be FinalizeAttachments event instead?

RFCs/0031-Multi-Test-Run-Finalization.md
```diff
@@ -0,0 +1,304 @@
...
+        /// </summary>
+        /// <param name="attachments">Finalised attachment sets.</param>
+        void HandleFinalisedAttachments(IEnumerable<AttachmentSet> attachments);
```
drognanar, line 101: nit: perhaps this should be called HandleFinalizedAttachmentsChunk? otherwise, the word finalized suggests that this could be the final set of attachments as well.

RFCs/0031-Multi-Test-Run-Finalization.md
```diff
@@ -0,0 +1,304 @@
...
+
+
+4. Use above logic to reprocess data collector attachments for parallel test executions and VS scenarios (e.g. `Run All Tests`, `Analyze Code Coverage for All Tests`). In case of `Analyze Code Coverage for All Tests` VS will use `vstest.console` in a variation of design mode and merge all code coverage reports. VS will show full code coverage report for all test projects.
+
+5. When [dotnet test](https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-test) command is used to execute unit tests in a given solution, new console app `Orchestrator` will be executed.
```
drognanar, line 116: will running the Orchestrator impact the performance of a run? or do we expect the same number of processes to be spawned?

RFCs/0031-Multi-Test-Run-Finalization.md
```diff
@@ -0,0 +1,304 @@
...
+  * `Orchestrator` for every project inside solution will start `dotnet test` command, using `Process` with output redirected. Output for every project will be printed to stdout with some mutex to not mix output from children.
```
drognanar, line 117: does this mean that `dotnet test` will run tests for different projects in parallel? will this be enabled by defaut?

RFCs/0031-Multi-Test-Run-Finalization.md
jakubch1: Created separate task. Its not changing interfaces so we can do it later.

RFCs/0031-Multi-Test-Run-Finalization.md
jakubch1: Created new task for it.

RFCs/0031-Multi-Test-Run-Finalization.md
jakubch1: Created new task for it.

RFCs/0031-Multi-Test-Run-Finalization.md
jakubch1: Renamed to: supportsIncrementalProcessing

RFCs/0031-Multi-Test-Run-Finalization.md
jakubch1: WDYT about moving 'dotnet test' specific things to separated document and keep it on opened PR? This document could focus only on finalization process, not so much on using it?

RFCs/0031-Multi-Test-Run-Finalization.md
jakubch1: Added this in arrows when starting processes

RFCs/0031-Multi-Test-Run-Finalization.md
jakubch1: @AbhitejJohn below my thinking, I can be totally wrong :)
1) For IDE before running finalization we know all test requests, so we know runsettings for projects. We will take all testadapter paths from all runsettings and we can call InitializeExtensions before running finalization
2) For commandline we will somehow figure out extensions paths (by getting through commandline or finding all runsettings for all projects). Can those paths be sent together with attachments and test statistics? 
When we have all paths we can always before running finalization call InitializeExtensions.

Another possibility is to add extension paths as parameter to FinalizeMultiTestRunAsync on VsTestConsoleWrapper.

RFCs/0031-Multi-Test-Run-Finalization.md
```diff
@@ -0,0 +1,304 @@
...
+
+        /// <summary>
+        /// Reprocess attachments generated by independent test executions
+        /// </summary>
+        /// <param name="attachments">Attachments to be processed</param>
+        /// <param name="progressReporter">Progress reporter. Accepts integers from 0 to 100</param>
+        /// <param name="logger">Message logger</param>
+        /// <param name="cancellationToken">Cancellation token</param>
+        /// <returns>Attachments after reprocessing</returns>
+        ICollection<AttachmentSet> ProcessAttachmentSets(ICollection<AttachmentSet> attachments, IProgress<int> progressReporter, IMessageLogger logger, CancellationToken cancellationToken);
```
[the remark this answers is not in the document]
jakubch1: good idea, changed it to return a Task

RFCs/0031-Multi-Test-Run-Finalization.md
```diff
@@ -0,0 +1,304 @@
...
+    public class MultiTestRunFinalizationProgressEventArgs : EventArgs
+    {
+        /// <summary>
+        /// Default constructor.
+        /// </summary>
+        /// <param name="currentProcessorIndex">Specifies current processor index.</param>
+        /// <param name="currentProcessorUris">Specifies current processor Uris.</param>
+        /// <param name="currentProcessorProgress">Specifies current processor progress.</param>
+        /// <param name="processorsCount">Specifies the overall number of processors.</param>
+        public MultiTestRunFinalizationProgressEventArgs(long currentProcessorIndex, ICollection<Uri> currentProcessorUris, long currentProcessorProgress, long processorsCount)
```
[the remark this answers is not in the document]
jakubch1: 👍  renamed

RFCs/0031-Multi-Test-Run-Finalization.md
```diff
@@ -0,0 +1,304 @@
...
+            CurrentProcessorUris = currentProcessorUris;
+            CurrentProcessorProgress = currentProcessorProgress;
+            ProcessorsCount = processorsCount;
+        }
+
+        /// <summary>
+        /// Gets a current processor index.
+        /// </summary>
+        [DataMember]
+        public long CurrentProcessorIndex { get; private set; }
```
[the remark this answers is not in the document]
jakubch1: Our code coverage report merger is merging files and removing not needed reports as DevOps is searching for all files with .coverage extension. Because of that we shouldn't run anything in parallel for same uris.

But in case we have two processors which are working on disjoint sets of Uris we can easily use them simultaneously

RFCs/0031-Multi-Test-Run-Finalization.md
```diff
@@ -0,0 +1,304 @@
...
+}
+```
+
+2. `MultiTestRunFinalizationProgressEventArgs` used by IMultiTestRunFinalizationEventsHandler:
+
+```
+namespace Microsoft.VisualStudio.TestPlatform.ObjectModel.Client
+{
+    [DataContract]
+    public class MultiTestRunFinalizationProgressEventArgs : EventArgs
```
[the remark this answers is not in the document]
jakubch1: Currently we pass IProgress<int> progressReporter to attachment processor. Every time reporter is used we propagate MultiTestRunFinalizationProgressEventArgs. When attachment processor is working we can't say which attachments are already done. We have separated callback to provide already processed attachments

RFCs/0031-Multi-Test-Run-Finalization.md
```diff
@@ -0,0 +1,304 @@
...
+        /// Collection of attachments.
+        /// </summary>
+        [DataMember]
+        public IEnumerable<AttachmentSet> Attachments { get; set; }
+
+        /// <summary>
+        ///  Gets or sets whether Metrics should be collected or not.
+        /// </summary>
+        [DataMember]
+        public bool CollectMetrics { get; set; }
```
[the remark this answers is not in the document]
jakubch1: I do not expect sending a lot of finalization requests. Currently tests execution is also sending it in every request.

RFCs/0031-Multi-Test-Run-Finalization.md
```diff
@@ -0,0 +1,304 @@
...
+
+If `IsAssociative` is `True` Test Platform may try to speed up whole process by reprocessing data collector attachments as soon as possible when any two test executions are done. For example let's assume we have 5 test executions which are generating 5 data collector attachments: `a1`, `a2`, `a3`, `a4` and `a5`. Test platform could perform invocations:
+* `var result1 = ProcessAttachmentSets([a1, a2, a3], ...);` when first 3 executions are done
+* `var result2 = ProcessAttachmentSets(result1.Concat([a4]), ...);` when 4th execution is done
+* `var finalResult = ProcessAttachmentSets(result2.Concat([a5]), ...);` when last test execution is done
+
+If `IsAssociative` is `False` then Test Platform will wait for all test executions to finish and call `ProcessAttachmentSets` only once:
+* `var finalResult = ProcessAttachmentSets([a1, a2, a3, a4, a5], ...);`
+
+By default `IsAssociative` should be `False`, unless processing can take longer time and it's beneficial to start the process as soon as possible.
```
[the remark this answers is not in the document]
jakubch1: multiTestRunCompleted  is exactly used for this purpose. TP will decide which attachment processors will be used based on this. multiTestRunCompleted  means that all test executions are done and we send to finalization all attachments.
@AbhitejJohn you were also proposing 'isLastBatch'. Do we want to rename it to 'isLastBatch'

I don't mind changing the naming here but I would like to agree finally on something as we have big PR in TP:
https://github.com/microsoft/vstest/pull/2463
and naming will hit it badly :)

Currently we use FinalizeMultiTestRun
Whole process is: Multi Test Run Finalization

@AbhitejJohn WDYT?

RFCs/0031-Multi-Test-Run-Finalization.md
```diff
@@ -0,0 +1,304 @@
...
+
+Introduce a new `IDataCollectorAttachmentProcessor` interface which can be implemented by Test Platform extensions and provide custom logic to reporcess (combine/merge) data collector attachments. Below interface should be used only for providing logic to reprocess information from data collector attachments from independent test executions. In other words data collector attachment should be modified/created using data from other data collector attachments. This interface should **not be** used for modifying any single data collector attachment. Test Platform will invoke `ProcessAttachmentSets` only if at least 2 data collector attachments related to processor (through `GetExtensionUris`) are created by independent test executions.
+
+```
+namespace Microsoft.VisualStudio.TestPlatform.ObjectModel.DataCollection
+{
+    /// <summary>
+    /// Interface for data collectors add-ins that choose to reprocess generated attachments
+    /// </summary>
+    public interface IDataCollectorAttachmentProcessor
```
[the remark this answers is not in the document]
jakubch1: this is part of vstest.console.exe process

RFCs/0031-Multi-Test-Run-Finalization.md
jakubch1: I've renamed it for SupportsIncrementalProcessing based on Abhitej's comment

I'm not sure if I understand correctly streaming API that you described. We could theoretically have such logic that you call:

FinalizeMultiTestRunAsync([a, b], isLastBatch: false,...); -> not returning any results keeps data in memory in TP, merges a, b
FinalizeMultiTestRunAsync([c, d], isLastBatch: false,...); -> not returning any results keeps data in memory in TP, merges previous results with c, d
FinalizeMultiTestRunAsync([e, f], isLastBatch: false,...); -> not returning any results keeps data in memory in TP, merges previous results with e,f
FinalizeMultiTestRunAsync([g, h], isLastBatch: TRUE,...); -> finally merges previous results with g,h and calls handlers callbacks

Such logic would be the simplest from usage point of view. As client does not need to keep any state on their side. It would complicate stuff inside TP. Interfaces needs to be modified in this case.

Keeping the state (already merged attachments) inside MultiTestRunFinalizationManager would allow to use same logic for parallel tests execution in TP. By this we could improve performance a lot. This would also not require sending back again partial results through translation layer.

from API perspective it could also be:

finalization = await vstestconsolewrapper.StartMultiTestRunFinalizationAsync(handler);
await finalization.ProcessAsync([a,b]); // provides progress info through handler
await  finalization.ProcessAsync([c,d]); // provides progress info through handler
await  finalization.ProcessAsync([e,f]); // provides progress info through handler
await  finalization.CloseAsync();    // provides back results through handler

@drognanar @AbhitejJohn  WDYT?

RFCs/0031-Multi-Test-Run-Finalization.md
jakubch1: if attachment processor supports incremental processing then invoking:

Process([pdb1, coverage1, pdb2, coverage2]) should give same results as
Process([Process([pdb1, coverage1, pdb2]), coverage2])  

There is no need to call ProcessAttachmentsSet([pdb1, coverage1, pdb2]) if ProcessAttachmentsSet([pdb1, coverage1, pdb2, coverage2]) was already done.

RFCs/0031-Multi-Test-Run-Finalization.md
jakubch1: Those are metrics used for telemetry. Currently we want to give back:
1) number of initial attachments
2) number of attachments after processing
3) final state of the process
4) time taken in seconds

Those metrics are not mergeable.

Added info to doc

RFCs/0031-Multi-Test-Run-Finalization.md
jakubch1: this is default currently for dotnet test on solution level. We don't want to change it

RFCs/0031-Multi-Test-Run-Finalization.md
jakubch1: currently all projects in solution ale executed in parallel which gives you n processes. Our change will add 2 additional processes: orchestrator and vstest.console for merging. All together we will have at least n+2 processes.

Every project in solution is starting additionally datacollector, testhost, and in case of CC also codecoverage.exe.
So in case of code coverage dotnet test we can have at least 4n + 2 processes. Maybe i still miss something

RFCs/0031-Multi-Test-Run-Finalization.md
jakubch1: Changed to:
        /// <summary>
        /// Dispatch ProcessedAttachmentsChunk event to listeners.
        /// </summary>
        /// <param name="attachments">Processed attachment sets.</param>
        void HandleProcessedAttachmentsChunk(IEnumerable<AttachmentSet> attachments);

RFCs/0031-Multi-Test-Run-Finalization.md
jakubch1: Changed to:    

    /// <summary>
        /// Dispatch ProcessedAttachmentsChunk event to listeners.
        /// </summary>
        /// <param name="attachments">Processed attachment sets.</param>
        void HandleProcessedAttachmentsChunk(IEnumerable<AttachmentSet> attachments);

RFCs/0031-Multi-Test-Run-Finalization.md
jakubch1: Added this comment to another task that we will extend this documenation

RFCs/0031-Multi-Test-Run-Finalization.md
jakubch1: fixed

RFCs/0031-Multi-Test-Run-Finalization.md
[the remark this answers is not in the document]
AbhitejJohn: The attachments may not be 'merged' from a processor standpoint. They could be split apart or additional metadata could be added or removed from them. A better term would be 'processed'. `ProcessTestRunAttachments` sounds like a good alternative to me given that the only functionality we are providing is exactly that - processing test run attachments across runs. I'm thinking the 'multi' part can be skipped given the scenario where processors could split/add/remove contents of a report. Given that, I'm not opposed to moving to `ProcessTestRunAttachments` if that makes sense to you as well.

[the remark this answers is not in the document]
AbhitejJohn: Functionally we should also make sure that the default code coverage processors do not throw if only one attachment is sent to them. They should ideally just return it back as is. I think you might already be doing so but just wanted to be explicit about it here.

RFCs/0031-Multi-Test-Run-Finalization.md
AbhitejJohn: So for IDE we already call InitializeExtensions based on the test adapters we find in the solution before a test discovery/execution. I'd recommend we reuse those. I'm assuming also that we are changing the finalization request to take in runsettings in which case the test platforms adapter search logic can be used to process that as well. 

The test platform also looks up extensions from the output directory if they aren't specified. I'd recommend using a file name pattern like we have for our other extensions *testadapter*.dll/*datacollector*.dll/*logger*.dll etc to easily find interesting extension assemblies. Perhaps *attachmentprocessor*.dll would work.

RFCs/0031-Multi-Test-Run-Finalization.md
AbhitejJohn: Sounds good to me

RFCs/0031-Multi-Test-Run-Finalization.md
jakubch1: I see in IVstestConsoleWrapper we have `string runSettings` for test runs and `string discoverySettings` for discovery. In case of finalization which runsettings we will use if we can several test projects with different runsettings. Maybe we could add parameter `IEnumerable<string> pathToAdditionalExtensions` to Finalize method?

RFCs/0031-Multi-Test-Run-Finalization.md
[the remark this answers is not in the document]
jakubch1: Regarding renaming we have now:

ProcessTestRunAttachments 

And whole process can be Test Run Attachments Processing or we can have Test Run Attachments Finalization

ITestRunAttachmentsProcessingEventsHandler  or ITestRunAttachmentsFinalizationEventsHandler

HandleTestRunAttachmentsProcessingComplete or HandleTestRunAttachmentsFinalizationComplete
HandleTestRunAttachmentsProcessingProgress or HandleTestRunAttachmentsFinalizationProgress

TestRunAttachmentsProcessingCompleteEventArgs or TestRunAttachmentsFinalizationCompleteEventArgs 
TestRunAttachmentsProcessingProgressEventArgs or TestRunAttachmentsFinalizationProgressEventArgs 

@AbhitejJohn @drognanar please reply with final proposal so I can do changes in TP :) I will consider writing simple script to go and rename everything :D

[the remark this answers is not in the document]
jakubch1: After more thinking about this maybe we could leave current naming: FinalizeMultiTestRun, MultiTestRunFinalization.
1) maybe in future we will want to merge/combine something else than attachments. Internally we can still use all functions and classes with current names.
2) finalize indicates that this is last thing of some process (test runs in this case). For someone who is reading our interface it could be easier to understand that this should be used after test runs

RFCs/0031-Multi-Test-Run-Finalization.md
jakubch1: Moved to PR: https://github.com/microsoft/vstest-docs/pull/236

RFCs/0031-Multi-Test-Run-Finalization.md
jakubch1: added processingSettings

RFCs/0031-Multi-Test-Run-Finalization.md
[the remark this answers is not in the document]
jakubch1: Renaming done

RFCs/0031-Multi-Test-Run-Finalization.md
jakubch1: As this part was removed, resolving it

RFCs/0031-Multi-Test-Run-Finalization.md
jakubch1: Discussed this with Abhitej. We will stick to current version.

RFCs/0031-Multi-Test-Run-Finalization.md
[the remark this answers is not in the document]
drognanar: just wondering if a single long will suffice, if we'd ever allow to run merges simultaneously, if that's not the case then having a single long is ok

RFCs/0031-Multi-Test-Run-Finalization.md
AbhitejJohn: >reporcess [](start = 62, length = 9)

spelling.

RFCs/0031-Multi-Test-Run-Finalization.md
```diff
@@ -0,0 +1,304 @@
+# 0031 Multi Test Run Finalization
+
+# Summary
+This document details a data collector extensibility point to reporcess (combine/merge) attachments obtained across test executions in both design mode and commandline scenarios.
+
+# Motivation
+Today when Test Platform executes tests in parallel only code coverage reports are merged (data collector attachments with uri: `datacollector://microsoft/CodeCoverage/2.0`). For other data collector attachments reprocessing is skipped and all of them are returned by Test Platform.
+
+The [dotnet test](https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-test) command is used to execute unit tests in a given solution. The `dotnet test` command builds the solution and runs a test host application for each test project in the solution. However, currently there is no way to reporcess (combine/merge) data collector attachments associated with each project execution. Code coverage reports are not merged.
```
AbhitejJohn: >reporcess [](start = 297, length = 9)

spelling here and in other place in this file. Perform a string replace.

RFCs/0031-Multi-Test-Run-Finalization.md
AbhitejJohn: >In other words data collector attachment should be modified/created using data from other data collector attachments [](start = 344, length = 116)

This isn't very clear.  Would rewording the following help:
`Below interface should be used only for providing logic to reprocess information from data collector attachments from independent test executions` -> `This interface should only be used to provider logic to reprocess information from data collector attachments from two or more test runs`

RFCs/0031-Multi-Test-Run-Finalization.md
jakubch1: fixed

RFCs/0031-Multi-Test-Run-Finalization.md
jakubch1: fixed

RFCs/0031-Multi-Test-Run-Finalization.md
jakubch1: this is some older sentences

PR #251, opened 2021-05-13
Add notice about branch rename to readme

nohwnd:


Sanan07:
Is it possible to add that pop-up message which appears in repo, when you open old branch? Or will it automatically work when default branch changes to main?

Sanan07 (approved):

nohwnd:
I already changed the main branch to *main*. There is no pop-up that I know of, but you should not land here unless you click a link that points you directly here (not much we could do about this unless we put notice in every article). If you go to the repo directly you will see the main *main* branch.

PR #262, opened 2022-06-10
Updating to include a required individual component

AbhitejJohn:
I've kept this to a Visual Studio requirement, how do we pull this additional dependency down when someone is using the commandline?

### DIFF
--- a/docs/contribute.md
+++ b/docs/contribute.md
@@ -14,8 +14,8 @@ Clone the repository to a local directory. Rest of this article assumes
 ```
 
 If you're planning to use **Visual Studio** as development environment, please
-install `VS 2017 RC 2` and the `.NET Core and Docker (Preview)` workload installed. See download
-link [here](https://www.microsoft.com/net/core#windowsvs2017).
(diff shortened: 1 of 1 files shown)


docs/contribute.md
```diff
@@ -14,8 +14,8 @@ Clone the repository to a local directory. Rest of this article assumes
 ```
 
 If you're planning to use **Visual Studio** as development environment, please
-install `VS 2017 RC 2` and the `.NET Core and Docker (Preview)` workload installed. See download
-link [here](https://www.microsoft.com/net/core#windowsvs2017).
+install `VS 2017 RC 2` or higher and the `.NET Core and Docker (Preview)` workload installed. See download
+link [here](https://www.microsoft.com/net/core#windowsvs2017). If you are using Visual Studio 2022 or higher, you would also need to install the inidividual component `.NET Portable Library targeting pack`.
```
nohwnd, line 18:
```suggestion
install `VS 2022`, and install individual component `.NET Portable Library targeting pack`.
```

